PR #43, opened 2020-02-25
Fix ApiError exception message handling

simu:
The ApiError exception does not require any custom code.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -1,0 +1,14 @@
+"""
+Unit-tests for helpers
+"""
+import commodore.helpers as helpers
+
+
+def test_apierror():
+    e = helpers.ApiError('test')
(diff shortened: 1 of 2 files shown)


tests/test_helpers.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def test_apierror():
+    e = helpers.ApiError('test')
+    assert(f"{e}" == 'test')
+
+    try:
+        raise helpers.ApiError('test2')
+    except helpers.ApiError as e2:
+        assert(f"{e2}" == 'test2')
```
bittner, line 14: You don't need braces for your `assert` statements. Those are not functions.

Examples:
```python
assert f"{e}" == "test"
assert f"{e2}" == "test2"
```

PR #44, opened 2020-02-25
Switch to Poetry

simu:
Switch to Poetry (https://python-poetry.org/) for managing development environment and building package.


srueg:
https://github.com/projectsyn/commodore/pull/102

PR #49, opened 2020-02-27
Allow empty facts

srueg:


commodore/cluster.py
```diff
@@ -59,7 +61,7 @@ def _full_target(cluster, components, catalog):
                 'catalog_url': f"{catalog}",
             },
             'cloud': {
-                'type': f"{cloud_provider}",
+                'provider': f"{cloud_provider}",
```
simu: Have you verified that all components expect and use `cloud.provider` rather than `cloud.type`?

srueg: Yes, and I changed the one occurrence I found.

simu:
> What's the reasoning behind allowing empty facts? Wouldn't it make more sense to require those basic facts to be available?
> 
> Writing components may get much more complicated if you cannot rely on those basic facts being available.

@srueg any inputs regarding this?

srueg:
Agree, it probably doesn't make sense to leave the cloud type and K8s distribution as optional. I'll change the code to explicitly fail. I'll leave the cloud region as optional though (as there might not be a region for every cloud).

commodore/cluster.py
```diff
@@ -43,14 +43,16 @@ def _full_target(cluster, components, catalog):
...
     component_defaults = [f"defaults.{cn}" for cn in components if
                           (P('inventory/classes/defaults') / f"{cn}.yml").is_file()]
+    global_defaults = ['global.common', f"global.{cluster_distro}", f"global.{cloud_provider}"]
+    if not cluster_distro:
+        raise click.ClickException(f"Required fact 'distribution' not set")
+    if not cloud_provider:
+        raise click.ClickException(f"Required fact 'cloud' not set")
+    if cloud_region:
+        global_defaults.append(f"global.{cloud_provider}.{cloud_region}")
+    global_defaults.append(f"{customer}.{cluster_id}")
```
bittner, line 53: There don't seem to be tests to ensure those changes. Please, don't forget to add them. This accumulates technical debt.

PR #50, opened 2020-02-28
Clean up catalog before writing refs

srueg:
Otherwise the refs will be cleaned up.

Signed-off-by: Simon Rüegg <simon.ruegg@vshn.ch>

### DIFF
--- a/commodore/catalog.py
+++ b/commodore/catalog.py
@@ -49,18 +49,23 @@ Compilation timestamp: {now}
 """
 
 
-def update_catalog(cfg, target_name, repo):
-    click.secho('Updating catalog repository...', bold=True)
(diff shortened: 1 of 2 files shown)


commodore/catalog.py
```diff
@@ -49,18 +49,22 @@ def _render_catalog_commit_msg(cfg):
 """
 
 
-def update_catalog(cfg, target_name, repo):
-    click.secho('Updating catalog repository...', bold=True)
-    # pylint: disable=import-outside-toplevel
-    from distutils import dir_util
-    import textwrap
+def clean_catalog(repo):
```
simu: Please add a status output when introducing "standalone" steps

PR #53, opened 2020-03-09
Allow reading API token from file

simu:
This commit changes Commodore to optionally read the API token from a
file. If the API token argument (command line or environment variable)
exists as a file, Commodore will read the file contents and use them as
the API token. Otherwise the value of the argument is used as the API
token. This allows using the service account's token when running
Commodore on Kubernetes.

Relates: SYN-21

### DIFF
--- a/commodore/config.py
+++ b/commodore/config.py
@@ -1,4 +1,5 @@
 from collections import namedtuple
+from pathlib import Path as P
 
 Component = namedtuple('Component', ['name', 'repo', 'version'])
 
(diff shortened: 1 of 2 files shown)


commodore/config.py
```diff
@@ -7,7 +8,13 @@ class Config:
     # pylint: disable=too-many-arguments
     def __init__(self, api_url, api_token, global_git, customer_git, verbose):
         self.api_url = api_url
-        self.api_token = api_token
+        self.api_token = None
+        if api_token is not None:
+            p = P(api_token)
```
srueg: Won't this raise an exception or anything if it's not a legal path (i.e. illegal characters in the token)?

simu: yep it does, when the token would contain a NULL byte :). I'll wrap it in some exception handling

simu: on the other hand, `click`'s argument handling breaks when the token contains a null byte,

simu: pathlib's Path does not choke on null bytes, I experimented a bit and arrived at the following minimal example: ```from pathlib import Path as P
test="test\u0000blubb"
print(len(test))
print(repr(test))
p = P(test)
print(p)
print(p.is_file())```

PR #80, opened 2020-05-07
Default to using Kapitan plugin on new components

srueg:
If this is forgotten, secrets won't be applied correctly and it silently
fails.


srueg:
I solved it in the ArgoCD component instead: https://git.vshn.net/syn/commodore-components/argocd/-/merge_requests/27

PR #88, opened 2020-05-12
Optional facts

srueg:
Change the CI setup to skip docker for tox and use GitHub actions matrix build.


commodore/git.py
```diff
@@ -161,10 +161,10 @@ def stage_all(repo):
                     # and a_blob as after.
                     before = c.b_blob.data_stream.read().decode('utf-8').split('\n')
                     after = c.a_blob.data_stream.read().decode('utf-8').split('\n')
-                    u = difflib.unified_diff(before, after, lineterm='',
-                                             fromfile=c.b_path, tofile=c.a_path)
-                    u = [_colorize_diff(l) for l in u]
-                    difftext.append('\n'.join(u).strip())
+                    diff = difflib.unified_diff(before, after, lineterm='',
```
simu, line 164: This change seems unrelated to commit 0de17c88863917826f6cd5af321040406ef07b5b

tests/test_target.py
```diff
@@ -0,0 +1,122 @@
...
+def test_render_target():
+    target = cluster._full_target(cluster_obj, components, catalog)
+    facts = cluster_obj['facts']
+    assert target != ""
+    all_classes = ([f"defaults.{cn}" for cn in components] +
+                   ['global.common',
+                       f"global.{facts['distribution']}",
+                       f"global.{facts['cloud']}",
+                       f"{cluster_obj['tenant']}.{cluster_obj['id']}",
+                    ])
```
bittner, line 31: Don't use braces here. This can backfire when some developer adds a comma in future. Also, the indentation seems to be out of sync in line 31.

tests/test_target.py
```diff
@@ -0,0 +1,122 @@
...
+    assert len(target['classes']) == len(
+        all_classes), "rendered target includes different amount of classes"
```
bittner, line 33: I'd suggest this style: (of course, personal taste)

```suggestion
    assert len(target['classes']) == len(all_classes), \
        "rendered target includes different amount of classes"
```

If you write it like this the actual test is on a single line. Note that the text afterwards, which is _totally optional_, is merely a comment and helps to understand the error in case the test fails.

srueg: Agree, autopep made this kinda uggly

srueg: See https://github.com/projectsyn/commodore/pull/89 for implementation of the findings

PR #89, opened 2020-05-12
Reformat tests

srueg:
Follow up for https://github.com/projectsyn/commodore/pull/88 review

### DIFF
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -1,0 +1,13 @@
+"""
+Unit-tests for git
+"""
+
+import commodore.git as git
+
+
+def test_reconstruct_api_response(tmp_path):
+    repo_path = tmp_path / 'test-repo'
+    repo_path.mkdir()
+    repo = git.create_repository(repo_path)
+    output = git.stage_all(repo)
+    assert output != ""
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -2,10 +2,11 @@
 Unit-tests for git
 """
 
-import commodore.git as git
+from commodore import git
+from pathlib import Path
 
 
-def test_reconstruct_api_response(tmp_path):
+def test_reconstruct_api_response(tmp_path: Path):
     repo_path = tmp_path / 'test-repo'
     repo_path.mkdir()
     repo = git.create_repository(repo_path)
--- a/tests/test_target.py
+++ b/tests/test_target.py
@@ -23,14 +23,13 @@ def test_render_target():
     target = cluster._full_target(cluster_obj, components, catalog)
     facts = cluster_obj['facts']
     assert target != ""
-    all_classes = ([f"defaults.{cn}" for cn in components] +
-                   ['global.common',
-                       f"global.{facts['distribution']}",
-                       f"global.{facts['cloud']}",
-                       f"{cluster_obj['tenant']}.{cluster_obj['id']}",
-                    ])
-    assert len(target['classes']) == len(
-        all_classes), "rendered target includes different amount of classes"
+    all_classes = [f"defaults.{cn}" for cn in components] + \
+        ['global.common',
+         f"global.{facts['distribution']}",
+         f"global.{facts['cloud']}",
+         f"{cluster_obj['tenant']}.{cluster_obj['id']}"]
+    assert len(target['classes']) == len(all_classes), \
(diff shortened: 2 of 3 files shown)


tests/test_git.py
```diff
@@ -0,0 +1,13 @@
+"""
+Unit-tests for git
+"""
+
```
bittner, line 4: I like to attach imports directly to the module docstring. Of course, this is personal taste.

tests/test_git.py
```diff
@@ -0,0 +1,13 @@
...
+import commodore.git as git
```
bittner, line 5:
```suggestion
from commodore import git
```

tests/test_git.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def test_reconstruct_api_response(tmp_path):
```
bittner, line 8: This is Python >= 3.6, correct? In this case I'd suggest:

```suggestion
def test_reconstruct_api_response(tmp_path: Path):
```

Doesn't really make the code safer, but we could run a linter over it to ensure that `tmp_path` is not any other type of object than `pathlib.Path`.

tests/test_git.py
```diff
@@ -0,0 +1,13 @@
...
+    repo_path = tmp_path / 'test-repo'
+    repo_path.mkdir()
+    repo = git.create_repository(repo_path)
+    output = git.stage_all(repo)
+    assert output != ""
```
bittner, line 13: I'm not sure I understand what's happening here. Do you:

1. create directory "test-repo"
1. run `git init` inside of it
1. stage the (empty) change set

That needs to be empty. True. :thinking:

tests/test_target.py
```diff
@@ -23,14 +23,13 @@ def test_render_target():
...
-                       f"global.{facts['cloud']}",
-                       f"{cluster_obj['tenant']}.{cluster_obj['id']}",
-                    ])
-    assert len(target['classes']) == len(
-        all_classes), "rendered target includes different amount of classes"
+    all_classes = [f"defaults.{cn}" for cn in components] + \
+        ['global.common',
+         f"global.{facts['distribution']}",
+         f"global.{facts['cloud']}",
+         f"{cluster_obj['tenant']}.{cluster_obj['id']}"]
```
bittner, line 30: I usually try to use the power of braces.

* They make the use of backslashes unneeded
* The code structure remains as we're used to function bodies in Java, JavaScript, CSS, etc.

```suggestion
    all_classes = [f"defaults.{cn}" for cn in components] + [
        'global.common',
        f"global.{facts['distribution']}",
        f"global.{facts['cloud']}",
        f"{cluster_obj['tenant']}.{cluster_obj['id']}",
    ]
```
Try to understand if you like this more. (I, personally, do prefer it this way.)

tests/test_target.py
```diff
@@ -23,14 +23,13 @@ def test_render_target():
...
+    assert len(target['classes']) == len(all_classes), \
+        "rendered target includes different amount of classes"
     # Test order of included classes
```
bittner, line 33: Does this need explanation? Maybe add it as a description (= second parameter) to the `assert` statement in the case?

(Otherwise, you'd need an empty line before the comment.)

tests/test_git.py
srueg: It's just a very basic test and makes sure no exceptions are thrown.

bittner: My point is, there's too much going on for a test. Just as some feedback.

bittner:
Fine. Feel free to merge.

tests/test_target.py
```diff
@@ -23,15 +23,13 @@ def test_render_target():
...
-                       f"{cluster_obj['tenant']}.{cluster_obj['id']}",
-                    ])
-    assert len(target['classes']) == len(
-        all_classes), "rendered target includes different amount of classes"
-    # Test order of included classes
+    all_classes = [f"defaults.{cn}" for cn in components] + [
+        'global.common',
+        f"global.{facts['distribution']}",
+        f"global.{facts['cloud']}",
+        f"{cluster_obj['tenant']}.{cluster_obj['id']}"]
```
bittner, line 30: You can probably move this assignment up, so all `assert`s are done after all initializations. That should make the code easier to read.

PR #91, opened 2020-05-13
Organize global classes in folders

srueg:
Add sub folders `distribution` and `cloud` for the respective classes.

Needs to be merged with:
* https://git.vshn.net/syn/commodore-defaults/-/merge_requests/6


srueg:
Updated docs: https://github.com/projectsyn/documentation/pull/43

PR #92, opened 2020-05-14
Update Various Dependencies

srueg:
Update the Kapitan docker image and various dependencies.

### DIFF
--- a/tools/Dockerfile.kapitan
+++ b/tools/Dockerfile.kapitan
@@ -1,9 +1,13 @@
-# Pushed to docker.io/vshn/kapitan:v0.26.0
+# Pushed to docker.io/vshn/kapitan:v0.27.2
 
 FROM docker.io/alpine:3.11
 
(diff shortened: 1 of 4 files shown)


tools/Dockerfile.kapitan
```diff
@@ -1,9 +1,13 @@
...
+ENV KAPITAN_VERSION=v0.27.2
 
-RUN apk add --no-cache curl \
+RUN apk add --no-cache \
+        ca-certificates \
+        curl \
  && curl -Lo /usr/local/bin/kapitan https://github.com/deepmind/kapitan/releases/download/${KAPITAN_VERSION}/kapitan-linux-amd64 \
  && chmod +x /usr/local/bin/kapitan
+
+USER 10001
```
simu, line 13: Is 10k+1 the correct user id here? Or did you mean 1001?

srueg: Doesn't really matter, as long as it's higher than 1000

PR #99, opened 2020-05-18
Create docker-compose and remove customer git fallback

srueg:
Simplify the user and SSH setup to make local usage easier.

By default the user's `.ssh/` directory with all keys gets mounted into the container. An SSH agent is started and the keys are added with `ssh-add` in the entrypoint. To skip the SSH agent setup, one can set the `SSH_AUTH_SOCK` env var to a path which is available in the container (mounted) to reuse an existing agent.

This will require some changes to the commodore job runner, MR is in progress.

Also, remove the customer git fallback.


corvus-ch:
When running `docker-compose run commodore compile $CLUSTER_ID`, the only thing I get is something along the line of `Agent pid 9`. I can not see whether or not commodore got executed and if there was a failure what it was.

srueg:
That might be because you're using a hardware token?

It executes `ssh-add`. In my case I get asked for the key password...

srueg:
> When running `docker-compose run commodore compile $CLUSTER_ID`, the only thing I get is something along the line of `Agent pid 9`. I can not see whether or not commodore got executed and if there was a failure what it was.

The issue was that the private key had a non-standard name (e.g. not `id_rsa`). We solved this by also adding non-standard keys in the entrypoint script and some error log for the case no keys were found.

PR #100, opened 2020-05-20
Component URLs

### LINKED ISSUE #73: Allow defining full path to Git repository for components
Enable Commodore to use components which are hosted in different repositories. This will likely require some reworking of the current component discovery mechanism, which relies on the `COMMODORE_GLOBAL_GIT_BASE` configuration to actually clone component repositories.

### Task deliverables
- [x] Commodore can clone components hosted on different Git services
- [x] A Git base URL is configured as a default for components
- [x] Overrides can be specified for single components in the config file (`commodore.yml`)
- [ ] In addition, component_versions can be used to change the URL for a component in the hierarchy

srueg:
Allow overwriting of component git repo URLs.
The component git repo URLs can be overwritten in a global Commodore
config file. The config file is searched for in the global inventory
repo. The global git base is used as default for non overwritten
components.

Closes #73

### DIFF
--- a/commodore/compile.py
+++ b/commodore/compile.py
@@ -60,24 +60,16 @@ def _regular_setup(config, cluster_id):
     customer_id = cluster['tenant']
 
     # Fetch components and config
-    try:
-        _fetch_global_config(config, cluster)
-        _fetch_customer_config(config, customer_id)
-        fetch_components(config)
-    except Exception as e:
-        raise click.ClickException(f"While cloning git repositories: {e}") from e
-
+    _fetch_global_config(config, cluster)
+    _fetch_customer_config(config, customer_id)
+    fetch_components(config)
     target_name = update_target(config, cluster)
     if target_name != 'cluster':
         raise click.ClickException(
             f"Only target with name 'cluster' is supported, got {target_name}")
 
     # Fetch catalog
-    try:
-        catalog_repo = fetch_customer_catalog(cluster['gitRepo'])
-    except Exception as e:
-        raise click.ClickException(
-            f"While cloning cluster catalog git repository: {e}") from e
+    catalog_repo = fetch_customer_catalog(cluster['gitRepo'])
 
     return cluster, target_name, catalog_repo
 
--- a/commodore/compile.py
+++ b/commodore/compile.py
@@ -47,7 +47,8 @@ def _fetch_customer_config(cfg, customer_id):
     if repopath is None:
         raise click.ClickException(
             f" > API did not return a repository URL for customer '{customer_id}'")
-    click.echo(f"Cloning {repopath}")
+    if cfg.debug:
+        click.echo(f" > Cloning customer config {repopath}")
     repo = git.clone_repository(repopath, P('inventory/classes') / customer_id)
     cfg.register_config('customer', repo)
 
@@ -69,7 +70,7 @@ def _regular_setup(config, cluster_id):
             f"Only target with name 'cluster' is supported, got {target_name}")
 
     # Fetch catalog
-    catalog_repo = fetch_customer_catalog(cluster['gitRepo'])
+    catalog_repo = fetch_customer_catalog(config, cluster['gitRepo'])
 
     return cluster, target_name, catalog_repo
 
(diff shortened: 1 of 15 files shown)


srueg:
For some reason the GitHub actions don't run. They're just stuck in `Starting your workflow run... ` state and then go to `This check was cancelled`

commodore/dependency_mgmt.py
```diff
@@ -8,6 +9,9 @@
 from .helpers import yaml_load
 
 
+Component = collections.namedtuple('Component', 'name repository_url target_directory')
```
simu, line 12: Consider reusing/extending `config.Component`, which is already used to track component Git URLs and desired versions.

commodore/dependency_mgmt.py
```diff
@@ -82,14 +77,49 @@ def fetch_components(cfg):
...
-        click.echo(f" > {c}...")
-        _fetch_component(cfg, c)
+        if cfg.debug:
+            click.echo(f" > Fetching component {c.name}...")
+        repo = git.clone_repository(c.repository_url, c.target_directory)
+        cfg.register_component(c.name, repo)
+        create_component_symlinks(cfg, c.name)
+
+
+def _read_component_urls(cfg, component_names):
```
simu, line 94: My personal preference (yes, I'm damaged by C) is to define methods above the point where they're called. I don't mind either way, but right now some methods used by `fetch_components` are defined above the method and this one is defined below.

commodore/compile.py
```diff
@@ -60,24 +62,16 @@ def _regular_setup(config, cluster_id):
...
 
     # Fetch components and config
-    try:
-        _fetch_global_config(config, cluster)
-        _fetch_customer_config(config, customer_id)
-        fetch_components(config)
-    except Exception as e:
-        raise click.ClickException(f"While cloning git repositories: {e}") from e
-
+    _fetch_global_config(config, cluster)
```
corvus-ch, line 65: Here you are removing the gift wrapping of exceptions. For the function `_relsymlink`. What is the thinking?

commodore/compile.py
```diff
@@ -115,7 +109,14 @@ def _local_setup(config, cluster_id):
             continue
         click.echo(f" > {c}")
         repo = git.init_repository(c)
-        config.register_component(c.name, repo)
+        component = Component(
+            name=c.name,
+            repo=repo,
+            version='master',
+            repo_url=None,
+            target_directory=P('dependencies') / c.name,
```
corvus-ch, line 117: I assume, that this will be the same for all instances. Consider adding some kind of a "constructor" that takes care of this.

commodore/compile.py
```diff
@@ -115,7 +109,14 @@ def _local_setup(config, cluster_id):
             continue
         click.echo(f" > {c}")
         repo = git.init_repository(c)
-        config.register_component(c.name, repo)
+        component = Component(
+            name=c.name,
+            repo=repo,
+            version='master',
+            repo_url=None,
```
simu, line 116: It would be possible to configure the repo_url here (probably a good idea to do so, in light of the pending local mode rewrite/improvements), by setting `repo_url=repo.remotes.origin.url`.

commodore/compile.py
srueg: The wrapping was to generic and added wrong/confusing error messages to unrelated exceptions

commodore/dependency_mgmt.py
```diff
@@ -65,12 +64,38 @@ def _discover_components(cfg, inventory_path):
...
+                f"{cfg.default_component_base}/{component_name}.git")
+        target_directory = P('dependencies') / component_name
+        component = Component(
+            name=component_name,
+            repo=None,
+            version='master',
+            repo_url=repository_url,
+            target_directory=target_directory,
+        )
+        components.append(component)
```
simu, line 97: *This comment is informational only, no changes required*

I did a short experiment with calling `cfg.register_component` here, and changing `fetch_components` to use `cfg.get_components()`, and an additional method `cfg.set_component_repo()`. However, the end result of that change does not really improve readability of the code, so I'm fine with leaving it like this for now.

commodore/compile.py
srueg: True. I'll create a property for this

PR #102, opened 2020-05-22
Switch to poetry

daiboruta:
Final switch to poetry with suggested changes


srueg:
Thanks for this PR! 🎉 

I rebased everything on current `master` and made some changes to the Dockerfile to improve the image size.

srueg:
I'm stuck on how to handle the version number. I though it should be possible by only relying on the `poetry version $VERSION` command and not have the version number hard coded (and adapted with `sed`). But now the `setup.py` can't access the version.

srueg:
I solved the issue by removing the `setup.py` since it's not used anymore with Poetry.

Dockerfile
```diff
@@ -1,60 +1,71 @@
...
 
-ARG BINARY_VERSION=unreleased
+COPY --from=helm_binding_builder \
+      /virtualenv/kapitan/inputs/helm/libtemplate.so \
+      /virtualenv/kapitan/inputs/helm/helm_binding.py \
+      /usr/local/lib/python3.8/site-packages/kapitan/inputs/helm/
 
-RUN sed -ie "s/^__version__ = 'Unreleased'$/__version__ = '$BINARY_VERSION'/" ./commodore/__init__.py
+COPY ./tools/entrypoint.sh /usr/local/bin/
+COPY ./lib/ /app/lib/
```
simu, line 63: It would be cleaner if building Commodore would package the `lib` and `filters` directory with the rest of the module.

I've tried this, and apart from poking at the search path in a couple locations in the code this can easily be done by moving the `lib` and `filters` directories under `commodore` and adding the files in `pyproject.toml` with an `include` statement. This change will then properly package the builtin jsonnet code in the wheel.

Not sure if this should be done as part of this PR or separately.

srueg:
Found an interesting blog post describing a comprehensive setup based on Poetry: https://cjolowicz.github.io/posts/hypermodern-python-01-setup/

.gitignore
```diff
@@ -50,6 +50,7 @@ pip-delete-this-directory.txt
 # Unit test / coverage reports
 htmlcov/
 .tox/
+.dtox/
```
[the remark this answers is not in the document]
daiboruta: I added that line because the .dtox folder got generated for me when I first downloaded the project. But it can be removed if it's not used

simu:
> I'm stuck on how to handle the version number. I though it should be possible by only relying on the `poetry version $VERSION` command and not have the version number hard coded (and adapted with `sed`). But now the `setup.py` can't access the version.

While the current implementation works when building tagged images, the GitHub action for building `commodore:latest` from the master branch will fail, as `latest` is not a valid version for a Python module, cf. https://python-poetry.org/docs/cli/#version and https://www.python.org/dev/peps/pep-0440/. Since my last comments, I've also figured out that we run `git describe` with `--abbrev=0` because otherwise `poetry build` fails due to the invalid Python module version.

I've tried out a number of different approaches, but if we want to continue having useful version information when executing `docker run --rm docker.io/vshn/commodore:latest --version`, we probably can't avoid having a call to `sed` in the `Dockerfile`.

Makefile
```diff
@@ -51,13 +51,13 @@ include tox.mk
 # Project parameters
 BINARY_NAME ?= commodore
 
-BINARY_VERSION = $(shell git describe --tags --always --dirty --match=v* || (echo "command failed $$?"; exit 1))
-VERSION ?= $(BINARY_VERSION)
+VERSION ?= $(shell git describe --tags --always --dirty=+dirty --abbrev=0 --match=v* || (echo "command failed $?"; exit 1))
```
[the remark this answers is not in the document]
simu: Reimplemented "Git version" in 055f222

Dockerfile
simu: Moved the files as part of this MR. Now building the Commodore Python module will produce a self-contained module which can be installed without requiring additional files to be added.

srueg:
Thanks again @daiboruta for kicking this off 👍

commodore/__init__.py
```diff
@@ -2,11 +2,9 @@
...
-__license__ = 'BSD'
 __url__ = 'https://git.vshn.net/syn/commodore/'
-# NOTE: This variable is set during the Docker image build process and *MUST*
-# be set somehow during the package build process if python package builds are
-# setup in the future
-__version__ = 'Unreleased'
+__git_version__ = "0"
+
+# provide Commodore installation dir as variable that can be imported
+from pathlib import Path as P
```
bittner, line 9: 1. Imports should be placed on top of the module. Pylint would complain.
1. What is the benefit of aliasing the `Path` with `P`? This seems to obscure things. Pylint, for example, complains about single-letter variables; it will likely complain about his, too.

simu: Since the pylint check passes, it looks like pylint is fine with the import location and the aliasing.

`Path` is aliased here, because it's currently aliased to `P` everywhere in the Commodore code base. I don't have strong feelings on whether we should refactor the aliasing everywhere or not.

PR #103, opened 2020-05-26
Handle empty facts

srueg:
If a fact is set to an empty string, handle it accordingly. Depending if
the fact is required or not, fail or ignore it respectively.

### DIFF
--- a/commodore/cluster.py
+++ b/commodore/cluster.py
@@ -46,7 +46,7 @@ def reconstruct_api_response(target_yml):
 def _full_target(cluster, components, catalog):
     cluster_facts = cluster['facts']
     for required_fact in ['distribution', 'cloud']:
-        if required_fact not in cluster_facts:
+        if required_fact not in cluster_facts or not cluster_facts[required_fact]:
             raise click.ClickException(f"Required fact '{required_fact}' not set")
 
     cluster_distro = cluster_facts['distribution']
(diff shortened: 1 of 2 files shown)


commodore/cluster.py
```diff
@@ -58,9 +58,9 @@ def _full_target(cluster, components, catalog):
     global_defaults = ['global.common',
                        f"global.distribution.{cluster_distro}",
                        f"global.cloud.{cloud_provider}"]
-    if 'region' in cluster_facts:
+    if 'region' in cluster_facts and cluster_facts['region']:
```
simu, line 61: It may make sense to ensure that the value of the optional fact is a string, and fail otherwise.

commodore/cluster.py
```diff
@@ -58,9 +58,9 @@ def _full_target(cluster, components, catalog):
...
         global_defaults.append(f"global.cloud.{cloud_provider}.{cluster_facts['region']}")
-    if 'lieutenant-instance' in cluster_facts:
+    if 'lieutenant-instance' in cluster_facts and cluster_facts['lieutenant-instance']:
```
simu, line 63: Same as above

commodore/cluster.py
srueg: This is already enforced by the API/K8s

simu: Never mind then :)

PR #104, opened 2020-05-26
Allow overriding of component URL in inventory

srueg:
In addition to the version this allows to override the Git URL for a component, using the `component_versions` dict. This allows for example the usecase of a forked component repository to be used for a certain part of the hierarchy.

Closes https://github.com/projectsyn/commodore/issues/73


srueg:
Tested with:
* Only overriding the URL (no version)
* Override url and version with branch
* Override url and version with tag
* Override the version (no url)

PR #106, opened 2020-05-28
Add helper to get proxy environment variables

corvus-ch:
Some environments require the `HTTP_PROXY, `HTTPS_PROXY` and `NO_PROXY`
variables being set. The helper `proxyVars` sets them when needed.

### DIFF
--- a/lib/commodore.libjsonnet
+++ b/lib/commodore.libjsonnet
@@ -1,3 +1,6 @@
+local kap = import 'lib/kapitan.libjsonnet';
+local params = kap.inventory().parameters;
+
 local namespaced(ns, obj) =
   obj {
     metadata+: { namespace: ns },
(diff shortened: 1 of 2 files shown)


corvus-ch:
> There's an argument of whether we should make `proxyVars` take `params` as an argument instead of importing `kapitan.libjsonnet` here […].

At first, I had it that way. Once I had done the n-th `local params = inv.parameters` and the m-th `com.envVars(params)`, I changed my mind.

lib/commodore.libjsonnet
```diff
@@ -55,13 +59,63 @@ local addNamespaceToHelmOutput(template_dir, namespace, exclude_objects=[]) =
...
+* In its simplest form, the object contains the keys `name` and `value`.
+*
+* This helper converts a dictionary into such a list where keys become `name`,
+* and values become `value`.
+*
+* \arg map Dictionary to be converted.
+*
+* \return List of dictionaries with `name`, `value` keys.
+*/
+local envList(map) = [
```
simu, line 106: I assume we need this for instances where we're not using `kube.libjsonnet`, as `kube.Container` supports setting `env_` and takes care of this conversion, see https://github.com/bitnami-labs/kube-libsonnet/blob/0e78b4332a893b7c1d03eefc84f9a2ef7d28b8ad/kube.libsonnet#L24-L40

corvus-ch: Exactly. E.g. when importing and manipulating an object from a static source or a Helm chart. I needed this in the ArgoCD component.

PR #110, opened 2020-06-11
Add Image for component testing

corvus-ch:
The linting image is meant as a runtime for linting Commodore
componentes.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -32,3 +32,18 @@ jobs:
         username: ${{ secrets.DOCKER_USERNAME }}
         password: ${{ secrets.DOCKER_PASSWORD }}
         buildargs: BINARY_VERSION
+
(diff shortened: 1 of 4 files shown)


.github/workflows/push.yml
```diff
@@ -32,3 +32,14 @@ jobs:
         username: ${{ secrets.DOCKER_USERNAME }}
         password: ${{ secrets.DOCKER_PASSWORD }}
         buildargs: BINARY_VERSION
+    - name: Build Linting Image
```
simu, line 35: Should be moved to its own job, we'll need to figure out if we really need/want to build linting images using Commodore version tags, or if just having commodore-lint:latest is enough.

This really could be a separate action.

corvus-ch: No we only need latest. At least for now. I limited to push to master branch.

Issue #72, opened 2020-04-07
Update Cookiecutter Component Template


srueg:
Update the [Cookiecutter template](https://github.com/projectsyn/commodore/tree/master/component-template) for new components to include the following:

- [x] A license: BSD-3-Clause
- [x] Option to base a component on a Helm chart (i.e. `--helm`) -> see  #118.
- [x] Add GitHub actions (jsonnet & yaml lint, others tbd)
- [x] `doc/` subfolder?
- [x] README.adoc template
- [x] CHANGELOG.adoc template
- [x] GitHub templates

How does the release process look like? Keyword: Changelog?

* Release process automation
* GitHub release preparation

tobru:
@corvus-ch Please do a real life experiment with https://github.com/appuio/component-openshift4-oauth/issues/5

corvus-ch:
The example in the openshift4-oauth component is done. It can be included in the cookiecutter.

Thoughts on how to improve:

GitHub strongly suggests to have a CONTRIBUTING.md file within each repository. For components we would (ideally) have the same file in each and every component repository. Instead of keeping them all in sync, we might opt to move that to somewhere at syn.tools and just place a link into CONTRIBUTING.md.

corvus-ch:
@srueg Can we extract the _option to base a component on a Helm chart_ to an issue on its own?

srueg:
Sure :+1:

corvus-ch:
In addition, also a code of conduct got added to the template.

I do not yet have an answer on how to do release automation.

tobru:
Done in #117

PR #117, opened 2020-06-17
Update Cookiecutter Component Template

corvus-ch:
This is the implementation for #72.


commodore/component_template.py
```diff
@@ -10,21 +11,27 @@
 from .helpers import yaml_load, yaml_dump
 
 
-def create_component(config, name, lib, pp):
+def create_component(config, name, lib, pp, owner, copyright):
+    slug = name.lower().replace(' ', '-')
```
corvus-ch, line 15: This is a duplication of the logic in `cookiecutter.json`. Can we reuse that somehow?

commodore/component_template.py
```diff
@@ -35,6 +42,8 @@ def create_component(config, name, lib, pp):
     git.add_remote(repo, 'origin', component.repo_url)
     index = repo.index
     index.add('*')
+    index.add('.github')
```
corvus-ch, line 45: Tried to use `.*`. This also added the `.git` folder. Gitception, yay!

component-template/{{ cookiecutter.slug }}/.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+  - name: ❓ Help and Support RocketChat Channel
```
corvus-ch, line 3: Do we need to make that configurable somehow?

commodore/component_template.py
simu: I guess we could reuse the result of cookiecutter, but we'd have to do the cookiecutter stuff first (in a temp dir to be on the safe side), and then check if the component already exists in `dependencies`.

component-template/{{ cookiecutter.slug }}/.github/ISSUE_TEMPLATE/config.yml
simu: maybe make the whole file optional (like we do the component library and postprocessing filter config)

component-template/{{ cookiecutter.slug }}/CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## [v0.1.0] - {{ cookiecutter.release_date }}
```
simu, line 9: Not sure I like having a `v0.1.0` in the template, but if we keep it, we should consider immediately tagging v0.1.0 on the initial commit.

commodore/cli.py
```diff
@@ -57,11 +57,16 @@ def compile(config, cluster, local, push, verbose):
               help='Add component library template')
 @click.option('--pp/--no-pp', default=False, show_default=True,
               help='Add component postprocessing template')
+@click.option('--owner', default="projectsyn", show_default=True,
+              help='The GitHub user or project name where the component will be hosted')
```
srueg, line 61:
```suggestion
              help='The GitHub user or organization name where the component will be hosted')
```

component-template/{{ cookiecutter.slug }}/.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
+
+Contributors guide: ./CONTRIBUTING.md
+-->
+
+## Checklist
+<!--
+Remove items that do not apply. For completed items, change [ ] to [x].
+-->
+
+- [ ] keep pull requests small so they can be easily reviewed.
```
srueg, line 13: Can we remove the periods in the list?

component-template/{{ cookiecutter.slug }}/CHANGELOG.md
srueg: Also not a big fan. Let's only keep `Unreleased`.

component-template/{{ cookiecutter.slug }}/Makefile
```diff
@@ -0,0 +1,34 @@
...
+.PHONY: lint
+lint: lint_jsonnet lint_yaml
+
+.PHONY: lint_jsonnet
+lint_jsonnet: $(JSONNET_FILES)
+	jsonnetfmt $(JSONNETFMT_ARGS) --test -- $?
+
+.PHONY: lint_yaml
+lint_yaml: $(YAML_FILES)
+	yamllint -f parsable -c $(YAMLLINT_CONFIG) $(YAMLLINT_ARGS) -- $?
```
srueg, line 27: I realised that if we'd use Docker in the Makefile, we could get rid of our custom CI image again and wouldn't require yamllint and jsonnet to be installed locally.
What do you think?

component-template/{{ cookiecutter.slug }}/playbook.yml
```diff
@@ -0,0 +1,36 @@
...
+  robots: disallow
+content:
+  sources:
+    - url: .
+      branches: HEAD
+      start_path: docs
+      edit_url: '{{ cookiecutter.github_url }}/edit/master/{path}'
+ui:
+  bundle:
+    url: https://github.com/vshn/antora-ui-default/releases/download/1.6/ui-bundle.zip
```
srueg, line 14:
```suggestion
    url: https://github.com/projectsyn/antora-ui-default/releases/download/1.3/ui-bundle.zip
```

component-template/{{ cookiecutter.slug }}/.github/PULL_REQUEST_TEMPLATE.md
corvus-ch: In ensured they all are proper sentences instead.

component-template/{{ cookiecutter.slug }}/Makefile
```diff
@@ -0,0 +1,38 @@
+MAKEFLAGS += --warn-undefined-variables
+SHELL := bash
+.SHELLFLAGS := -eu -o pipefail -c
+.DEFAULT_GOAL := all
+.DELETE_ON_ERROR:
+.SUFFIXES:
+
+DOCKER_CMD   ?= docker
+DOCKER_IMAGE ?= docker.io/projectsyn/commodore-component-test:latest
```
srueg, line 9: Let's use the official upstream images of the tools, so we can get rid of our custom image again.

component-template/{{ cookiecutter.slug }}/CHANGELOG.md
corvus-ch: I tried to find a way to showcase how the change log is supposed to be used. My assumption is, if left out, the change log will become everything but consistent. Also I was unable to find a compare link comparing the first commit with HEAD without knowing the commit ID of that first commit. Immediately tagging that first commit, seems to be the cleanest option. Probably will go with `v0.0.1` instead.

component-template/{{ cookiecutter.slug }}/.github/ISSUE_TEMPLATE/config.yml
corvus-ch: I would like to always have the `blank_issues_enabled: false`. Removing the file, as we do it with the others, does not work here. But this is easy enough in Jinja. The question remains, shall we bother?

commodore/component_template.py
corvus-ch: Hmm, I like that idea.

component-template/{{ cookiecutter.slug }}/.github/ISSUE_TEMPLATE/config.yml
corvus-ch: We keep that for the moment. We will change that once we have the need for it.

component-template/{{ cookiecutter.slug }}/CHANGELOG.md
simu: I like v0.0.1 for the initial commit much better :)

corvus-ch: Discussed this this morning in our daily. We agreed on removing that initial version keeping only the _Unrelease_. We also keep the link for unreleased using the tag `v0.1.0` as the comparison base. We accept the fact, that this will be an invalid link until a first version will be released.

commodore/component_template.py
corvus-ch: We need to come to an end here. I will create a ticket for that to be handled later.

component-template/{{ cookiecutter.slug }}/Makefile
```diff
@@ -0,0 +1,41 @@
+MAKEFLAGS += --warn-undefined-variables
+SHELL := bash
+.SHELLFLAGS := -eu -o pipefail -c
+.DEFAULT_GOAL := all
+.DELETE_ON_ERROR:
+.SUFFIXES:
+
+DOCKER_CMD   ?= docker
+DOCKER_ARGS  ?= run -v "$${PWD}:/component" --workdir /component
```
srueg, line 9:
```suggestion
DOCKER_ARGS  ?= run --rm --user "$$(id -u)" -v "$${PWD}:/component" --workdir /component
```

component-template/{{ cookiecutter.slug }}/Makefile
```diff
@@ -0,0 +1,41 @@
...
+.PHONY: lint_yaml
+lint_yaml: $(YAML_FILES)
+	$(YAMLLINT_DOCKER) -f parsable -c $(YAMLLINT_CONFIG) $(YAMLLINT_ARGS) -- $?
+
+.PHONY: format
+format: format_jsonnet
+
+.PHONY: format_jsonnet
+format_jsonnet: $(JSONNET_FILES)
+	$(JSONNET_DOCKER) -- chmod +x $$(command -v jsonnetfmt) && jsonnetfmt $(JSONNETFMT_ARGS) -- $?
```
srueg, line 41: Why do we need the `chmod +x` here?

corvus-ch: Oh, that is a leftover.

```suggestion
	$(JSONNET_DOCKER) $(JSONNETFMT_ARGS) -- $?
```

component-template/{{ cookiecutter.slug }}/Makefile
```diff
@@ -0,0 +1,41 @@
...
+SHELL := bash
+.SHELLFLAGS := -eu -o pipefail -c
+.DEFAULT_GOAL := all
+.DELETE_ON_ERROR:
+.SUFFIXES:
+
+DOCKER_CMD   ?= docker
+DOCKER_ARGS  ?= run -v "$${PWD}:/component" --workdir /component
+
+JSONNET_FILES   ?= $(shell find . -type f -name '*.*jsonnet' -or -name '*.libsonnet')
```
srueg, line 11: Looks like this breaks if the files/folders contain spaces?
If I run `make` I get the following error:
```
make: *** No rule to make target `lib/{{', needed by `lint_jsonnet'.  Stop.
```

PR #122, opened 2020-06-24
Render Single Component

### LINKED ISSUE #74: Support rendering manifests for a single component
To simplify developing components it should be possible to have Commodore render just the manifests generated by that component, ideally without requiring any external APIs. This will also be a useful building block for implementing automated validation and testing of components.

Potentially the changes required for #71 can be reused to implement this feature.

### Task deliverables

- [x] Manifests generated by a single component can be rendered by Commodore
- [x] No pre-setup workspace or external API interaction is required to render a component's manifests

srueg:
Render a single component by faking the required parts of the inventory. Additional classes can be specified with the `--values/-f` flag (multiple possible).
For the compilation to succeed, a user is responsible to provide all required parameters (which are not covered in the `defaults.yml`).

This changes the CLI structure: There's two new command groups: `catalog` and `component`. The previous `compile` and `clean` commands are moved to the `catalog` group and the `new-component` was moved together with the `compile-component` command into the `component` group:

```commodore component new```
```commodore component compile```
```commodore catalog compile```

~~This first implementation doesn't support secret refs. This means all used secret refs in the `defaults.yml` must be overridden with concrete values in the provided classes.~~
Vault secret refs are supported and get revealed to the path they reference (e.g. `?{vaultkv:${customer:name}/${cluster:name}/backup-cluster-objects/password}` becomes `t-silent-test-1234/c-green-test-1234/backup-cluster-objects/password`)

Closes #74

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+MAKEFLAGS += -j4
+
 pages   := $(shell find docs -type f -name '*.adoc')
 web_dir := ./_antora
 
@@ -63,10 +65,3 @@ docker:
 		--build-arg GITVERSION=$(GITVERSION) \
 		-t $(IMAGE_NAME) .
 	@echo built image $(IMAGE_NAME)
-
-COMPONENT_TEST_IMAGE_NAME ?= docker.io/vshn/$(BINARY_NAME)-component-test:$(VERSION)
-
-.PHONY: docker-component-test
-docker-component-test:
-	docker build -t $(COMPONENT_TEST_IMAGE_NAME) -f tools/Dockerfile.component_test .
-	@echo built image $(COMPONENT_TEST_IMAGE_NAME)
--- a/commodore/component/compile.py
+++ b/commodore/component/compile.py
@@ -34,30 +34,8 @@ def compile_component(config: Config, component_path, value_files, search_paths,
     try:
         if config.debug:
             click.echo(f"   > Created temp workspace: {temp_dir}")
-        for d in ['classes/components', 'classes/defaults', 'targets']:
-            os.makedirs(temp_dir / 'inventory' / d, exist_ok=True)
-        dependencies_path = temp_dir / 'dependencies'
-        dependencies_path.mkdir(exist_ok=True)
-        # Create class symlink
-        relsymlink(
-            component_class_file.parent,
-            component_class_file.name,
-            temp_dir / 'inventory/classes/components')
-        # Create defaults symlink
-        relsymlink(
-            component_defaults_file.parent,
-            component_defaults_file.name,
-            temp_dir / 'inventory/classes/defaults',
-            f"{component_name}.yml")
-        # Create component symlink
-        relsymlink(
-            component_path.parent,
-            component_path.name,
-            dependencies_path,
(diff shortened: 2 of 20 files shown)

### CODE REVIEW

**Comment by srueg** [comment]
Not done yet / Caveats:
* The `argocd.libjsonnet` library is being faked with an empty object, therefore the Argo CD `application` manifest won't be rendered
* No jsonnet lib dependencies are created, therefore the `kube.libjsonnet` is missing (which breaks many components)

**Comment by srueg** [comment]
I fixed the remaining open points and was even able to find a solution to support vault references: By creating a fake VaultBackend, references are "revealed" to the path in Vault they specify.

File: commodore/component/compile.py [hunk 1]
```diff
@@ -0,0 +1,140 @@
+import os
+from pathlib import Path as P
+import shutil
+import tempfile
+
+import click
+
+from commodore.config import Config
+from commodore.dependency_mgmt import fetch_jsonnet_libs, _relsymlink
```
simu, line 9: Consider refactoring `_relsymlink` into `helpers.py` and making it "public".

commodore/component/compile.py
```diff
@@ -0,0 +1,140 @@
...
+        if config.trace:
+            click.echo(f" > Temp dir left in place {temp_dir}")
+        else:
+            if config.debug:
+                click.echo(f" > Remove temp dir {temp_dir}")
+            shutil.rmtree(temp_dir)
+
+
+def _prepare_fake_inventory(temp_dir: P, component_name, component_path, value_files):
+    component_class_file = component_path / f"class/{component_name}.yml"
```
simu, line 108:
```suggestion
    component_class_file = component_path / 'class' / f"{component_name}.yml"
```

commodore/component/compile.py
```diff
@@ -0,0 +1,140 @@
...
+    component_defaults_file = component_path / 'class/defaults.yml'
```
simu, line 109:
```suggestion
    component_defaults_file = component_path / 'class' / 'defaults.yml'
```

srueg:
I moved as many changes into own PRs as possible. This last one still includes the change of command structure and the actual component compile functionality since they're too closely intertwined.
PR #130 needs to be merged first since it switches to using the Kapitan library instead of running the binary.

Makefile
```diff
@@ -63,10 +65,3 @@ docker:
 		--build-arg GITVERSION=$(GITVERSION) \
 		-t $(IMAGE_NAME) .
 	@echo built image $(IMAGE_NAME)
-
-COMPONENT_TEST_IMAGE_NAME ?= docker.io/vshn/$(BINARY_NAME)-component-test:$(VERSION)
-
-.PHONY: docker-component-test
-docker-component-test:
-	docker build -t $(COMPONENT_TEST_IMAGE_NAME) -f tools/Dockerfile.component_test .
-	@echo built image $(COMPONENT_TEST_IMAGE_NAME)
```
simu, line 72: This change doesn't need to be in this PR anymore, this already has been removed in #127

corvus-ch: Sadly we missed that one in #127 so it has to stay.

PR #124, opened 2020-06-29
Mark `commodore.filterNull()` deprecated

corvus-ch:


commodore/lib/commodore.libjsonnet
```diff
@@ -6,6 +6,13 @@ local namespaced(ns, obj) =
...
   };
 
+/**
+ * \brief A function to filter NULL values.
+ *
+ * \arg `list` The array to filter
+ *
+ * \deprecated Use `std.prune(a)` instead.
+ */
 local filterNull(list) = std.filter(function(obj) obj != null, list);
```
simu, line 16: I would also produce the deprecation warning when the function is called, as shown below.

```suggestion
local filterNull(list) = std.trace('filterNull is deprecated, use std.prune(a) instead', std.filter(function(obj) obj != null, list));
```

PR #140, opened 2020-07-09
Allow overriding Git user info for catalog commits

daiboruta:
This commit introduces new arguments for `commodore catalog compile` for
overriding local Git user information when pushing commits to a cluster
catalog.

By default Commodore will use the local Git user information for catalog
commits.
To improve compatibility with existing Git configurations, the arguments
are named `--git-author-name` and `--git-author-email` to pick up the
familiar `GIT_AUTHOR_NAME` and `GIT_AUTHOR_EMAIL` environment variables
without any further configuration.

Please note: Commodore does not allow you to override commit author and
committer separately, and will use the provided author name and email,
if any, as both commit author and committer.

The commit also updates docker-compose.yaml to map `~/.gitconfig` into
the container to propagate the global Git config into the container.
This simplifies reusing the Git author configuration on the host system
when running Commodore locally using docker-compose.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/cli.py
+++ b/commodore/cli.py
@@ -61,16 +61,23 @@ def clean(config: Config, verbose):
                     'the Lieutenant API or fetch/push Git repositories.'))
 @click.option('--push', is_flag=True, default=False,
               help='Push catalog to remote repository.')
+@click.option('--git-author-name', envvar='GIT_AUTHOR_NAME', metavar='USERNAME', help='Name of catalog commit author')
+@click.option('--git-author-email', envvar='GIT_AUTHOR_EMAIL', metavar='EMAIL',
+              help='E-mail address of catalog commit author')
 @verbosity
 @pass_config
 # pylint: disable=too-many-arguments
-def compile_catalog(config: Config, cluster, api_url, api_token, global_git_base, local, push, verbose):
+def compile_catalog(config: Config, cluster, api_url, api_token,
+                    global_git_base, local, push, verbose, git_author_name,
+                    git_author_email):
     config.update_verbosity(verbose)
     config.api_url = api_url
     config.api_token = api_token
     config.global_git_base = global_git_base
     config.local = local
     config.push = push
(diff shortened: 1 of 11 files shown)


srueg:
Thanks for this PR @daiboruta 
Can you please rebase your changes on the latest `master` and resolve the conflicts?

commodore/config.py
```diff
@@ -34,6 +34,12 @@ def __init__(self, api_url, api_token, global_git, verbose):
         self._components = {}
         self._config_repos = {}
         self._verbose = verbose
+        
+        self.username = None
```
srueg, line 38: These assignements to `None` and the following `if is not None` seems unnecessary, since the default for these parameters is `None` anyway.

commodore/cli.py
```diff
@@ -26,12 +26,14 @@ def _version():
 @click.option('--api-token', metavar='TOKEN', help='Lieutenant API token')
 @click.option('--global-git-base', metavar='URL',
               help='Base directory for global Git config repositories')
+@click.option('--username', metavar='USERNAME', help='Username used to make the commits')
```
srueg, line 29:
```suggestion
@click.option('--username', metavar='USERNAME', help='User name used to make the commits')
```

commodore/cli.py
```diff
@@ -61,16 +61,23 @@ def clean(config: Config, verbose):
                     'the Lieutenant API or fetch/push Git repositories.'))
 @click.option('--push', is_flag=True, default=False,
               help='Push catalog to remote repository.')
+@click.option('--git-author-name', envvar='GIT_AUTHOR_NAME', metavar='USERNAME', help='Name of catalog commit author')
```
srueg, line 64: Should we add default values here? I think currently a name & mail is required without any default/fallback.

srueg: Never mind, the default is the local Git info.

PR #148, opened 2020-07-24
Fix read token from file

srueg:
Always try to read the token from a file, not only in the constructor.

## Checklist

### DIFF
--- a/commodore/config.py
+++ b/commodore/config.py
@@ -21,16 +21,7 @@ class Config:
     def __init__(self, api_url=None, api_token=None, global_git=None, verbose=False, username=None, usermail=None):
         self.api_url = api_url
         self.api_token = None
-        if api_token is not None:
-            try:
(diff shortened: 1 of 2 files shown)


commodore/config.py
```diff
@@ -59,6 +50,19 @@ def config_file(self):
...
+        return self.__api_token
+
+    @api_token.setter
+    def api_token(self, api_token):
+        if api_token is not None:
+            p = P(api_token)
+            if p.is_file():
+                with open(p) as apitoken:
+                    api_token = apitoken.read()
+            self.__api_token = api_token.strip()
```
simu, line 64: Usually a single underscore is used to mark a field "private", double underscores are handled specially to allow "class private" variables, cf. https://docs.python.org/3/tutorial/classes.html#private-variables

Here, a single underscore should be sufficient

PR #149, opened 2020-07-27
Add folders generated during catalog compilation

akosma:
This PR includes folders that were previously ignored. 

## Reason

The reason is the `docker run projectsyn/commodore catalog compile` command which generates those folders as belonging to the root user, which prevents from writing to them. Situation found on a Ubuntu 18.04 system. The solution was to `mkdir` the folders before running the command.

## Checklist

- [ ] Update tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 _antora
 
 ## Ignore commodore output
+catalog
 compiled
 dependencies
 inventory
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@ _antora
 
 ## Ignore commodore output
 compiled
+dependencies
+inventory
 ## From here: GitHub's gitignore template for Python
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/.gitignore
+++ b/.gitignore
@@ -2,10 +2,7 @@
 _antora
 
 ## Ignore commodore output
-catalog
 compiled
-dependencies
-inventory
 ## From here: GitHub's gitignore template for Python
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/README.md
+++ b/README.md
@@ -132,6 +132,8 @@ using `helm template`.
 
 ## Run Commodore in Docker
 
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -2,10 +2,7 @@
 _antora
 
 ## Ignore commodore output
-catalog
```
srueg, line 5: Can we leave them ignored? Otherwise all compiled changes and the inventory would be tracked by git.

simu:
This won't work as is, because Commodore will delete the folders completely when running `commodore catalog clean` and recreate them without the `.gitkeep` files.

akosma:
I forgot to mention in the description the following key point: the problem happens when running Commodore using Docker, the folders are created as belonging to `root` and not to the current user. This happens even if the `--user` parameter is given a correct uid. In this case, this is the error message:

```
Agent pid 10
Enter passphrase for /app/.ssh/id_ed25519:
Identity added: /app/.ssh/id_ed25519 (adrian.kosmaczewski@vshn.ch)
Cleaning working tree
Updating global config...
Error: While cloning git repository: Cmd('git') failed due to: exit code(128)
  cmdline: git clone -v ssh://git@github.com/projectsyn/commodore-defaults.git inventory/classes/global
  stderr: 'fatal: could not create leading directories of 'inventory/classes/global': Permission denied
```

Any ideas? The current workaround involves a `mkdir catalog inventory dependencies` before running the `docker run` command.

srueg:
Seems like adding `mkdir -p catalog inventory dependencies` to the README is the only solution?

akosma:
> Seems like adding `mkdir -p catalog inventory dependencies` to the README is the only solution?

Yes, maybe that's for now the only way forward. I'll also make this very explicit in the "quick started" tutorial I'm writing.

akosma:
I've reverted all changes and added a note in the README for this case.

simu:
I had a look at this, and the problem is that if the folders don't exist on the host, **docker** will create them when setting up the bind mounts and as far as I've been able to tell there's no way to tell docker to use a specific user id when creating host folders for bind mounts.

PR #155, opened 2020-07-31
Allow postprocessing filters to be disabled

simu:
It can be helpful to make postprocessing components optional based on a reclass reference.

This PR implements this feature by introducing an optional field `enabled` for each filter definition. If the field doesn't exist, filters will count as enabled. This feature is particularly useful if we want to apply a filter to a subchart of a Helm chart which can be enabled or disabled, e.g. Postgres for Keycloak.

The PR also introduces a first set of tests for `postprocess_components`.

## Checklist

- [x] Update tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 * Validation of component slug ([#153])
 * `component compile` now applies postprocessing filters ([#154])
+* Option to disable postprocessing filters ([#155])
 
 ## [v0.2.2]
 
@@ -145,3 +146,4 @@ Initial implementation
 [#151]: https://github.com/projectsyn/commodore/pull/151
 [#153]: https://github.com/projectsyn/commodore/pull/153
 [#154]: https://github.com/projectsyn/commodore/pull/154
+[#155]: https://github.com/projectsyn/commodore/pull/155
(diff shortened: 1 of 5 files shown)


simu:
Rebased on latest master after merging #154

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 * Validation of component slug ([#153])
 * `component compile` now applies postprocessing filters ([#154])
+* Added option to disable postprocessing filters ([#155])
```
srueg, line 14:
```suggestion
* Option to disable postprocessing filters ([#155])
```

commodore/postprocess/__init__.py
```diff
@@ -8,6 +8,17 @@
 from .builtin_filters import run_builtin_filter
 
 
+def _inventory_resolve(inventory, value):
```
srueg, line 11: Seems kinda wrong to implement this here. But I couldn't find a way to reuse the reclass implementation quickly.

simu: I thought about this a bit after seeing the comment. What I've come up with is that we could probably move the whole postprocessing filters definition into the inventory as `parameters.commodore.postprocess.filters` in the component class (similarly to the Kapitan compilation definitions).

That way we'd get already resolved filter arguments etc.

I'd suggest opening an issue for this refactoring, and keep the manual inventory resolving for this PR.

simu: After looking at the code again I realized that we already have a partial implementation to resolve references for the builtin filters' `filterargs`. I've refactored the postprocessing code a bit to use this implementation for both filterargs and any other fields in the filter definition dicts.

PR #147, opened 2020-07-23
Rework Documentation, part 1

simu:
This is a first PR working towards completing #120.

This PR contains the following documentation pages: 
* A list of all commands provided by commodore with short explanations of what they do
* A list of all options for all commands (manpage-inspired)
* A page documenting Commodore concepts
* A page documenting Commodore architecture

## Checklist

- [x] Keep pull requests small so they can be easily reviewed.
- [x] Update the documentation.
- [x] Update the ./CHANGELOG.md.
- [x] Link this PR to related issues.

PR #160, opened 2020-08-12
Component Discovery With Reclass

### LINKED ISSUE #138: Fake Components During Discovery
## Context
During discovery, all components are being cloned, even ones that are not included for the current cluster. Faking the components in this step could improve performance a lot. Additionally would it lower the side effects components can have to each other. 
After the inventory is built, the actually included components of a cluster can be cloned.

srueg:
One consequence of this is, that the hierarchy can't reference parameters which are defined in component classes (i.e. defaults). This is due to the fact that a reclass inventory must be able to resolve all references in order to be rendered. There's no mechanism to ignore invalid references (like there is for ignoring inexistent class files). 

## Checklist

Closes #138

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 * Reworked documentation ([#147])
+* Use Reclass to discover components ([#160])
 
 ## [v0.2.2]
 
@@ -159,4 +160,5 @@ Initial implementation
 [#155]: https://github.com/projectsyn/commodore/pull/155
 [#157]: https://github.com/projectsyn/commodore/pull/157
 [#158]: https://github.com/projectsyn/commodore/pull/158
+[#160]: https://github.com/projectsyn/commodore/pull/160
 [#161]: https://github.com/projectsyn/commodore/pull/161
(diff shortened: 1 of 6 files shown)


Kidswiss:
As from today's sync:

Tested to access a puppet profile variable in hiera, it doesn't seem to be working.

corvus-ch (approved):

simu (approved):

tobru:
Please also make sure that this is documented, maybe needs to be done after #147 has been merged.

srueg:
As discussed during the SYNc today: Since this is also not allowed in Puppet and it's probably better that way anyway, we'll proceed with this approach and from now on don't allow parameters which are set in components to be referenced from the inventory.

simu:
Currently testing this for some "common" usecases. What I found so far is that with this approach we'll have to rethink how local mode works, as you now can't just add a component in the inventory somewhere and run `catalog compile --local` to inspect the changes.

srueg:
I don't think I understand the problem. This wasn't possible before either?

simu:
It is currently possible for any component that the file-based discovery mechanism finds and downloads.

simu (approved):

tobru (approved):

PR #169, opened 2020-09-04
Rename component template GitHub workflow config to test.yml

simu:
Apparently GitHub won't pick up workflows defined in files with extension .yaml.

## Checklist


srueg:
I don't think that's the issue. As from the documentation:
> Workflow files use YAML syntax, and must have either a .yml or .yaml file extension
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#about-yaml-syntax-for-workflows

simu:
Well, for sure the actions did not work when the file was named `test.yaml`, and started working immediately when I renamed it (e.g. on projectsyn/component-fluentbit)

simu:
Hmm, looks like it may be unrelated to the file name but it appears that GitHub only detects the file after it's renamed or something, cf. https://github.community/t/no-github-actions-workflows-detected-in-repo/16085

simu:
The file name is not the issue for workflows sometimes not being picked up. Closing this PR.

PR #176, opened 2020-09-10
Update default Vault values

srueg:
## Checklist


simu (approved):
LGTM. You probably want to update the default back-end name in https://github.com/projectsyn/commodore-defaults/blob/1226e0c5de322b7f28912fdd9691da2dd4068830/common.yml#L4 as well

PR #178, opened 2020-09-14
Add "catalog list" command

### LINKED ISSUE #135: Safeguard "clean" Command
## Context
The commodore clean command removes the entire inventory and all components. Changes which are not checked into git will be lost. On a regular catalog compile (if not `--local` mode), the clean command is run first.

The safeguard should interactively ask if changes should really be dropped in case of a dirty workspace. A `--force` flag could be implemented to drop changes without asking first (for example in a CI/CD or test context).

chrisglass:
This commit adds the "commodore catalog list" command that will query
lieutenant for a list of available cluster IDs (and descriptions), and
display the information on stdout.

This partially fixes #135 

## Checklist

- [ ] Update tests.

## Testing done

Unit tests have not been added since the logic is trivial (basically printing out contents of an already-existing API) and consists mostly of stiching side-effects. Mocking side-effects (the click library and lieutenant function call) would result it unit testing a for loop, which I suspect is not very interesting.


chrisglass:
Hmm, I seem to have messed up my commit stack :)

Fixing.

PR #179, opened 2020-09-14
Add "catalog list" command

### LINKED ISSUE #135: Safeguard "clean" Command

chrisglass:
This commit adds the "commodore catalog list" command that will query
lieutenant for a list of available cluster IDs (and descriptions), and
display the information on stdout.

This partially helps with #135 

## Checklist

- [ ] Update tests.

## Testing done

Unit tests have not been added since the logic is trivial (basically printing out contents of an already-existing API) and consists mostly of stiching side-effects. Mocking side-effects (the click library and lieutenant function call) would result it unit testing a for loop, which I suspect is not very interesting.

### DIFF
--- a/docs/modules/ROOT/pages/reference/cli.adoc
+++ b/docs/modules/ROOT/pages/reference/cli.adoc
@@ -92,6 +92,10 @@ See also xref:local-mode.adoc[local mode documentation].
 *--help*::
   Show catalog clean usage and options then exit.
 
+== Catalog List
+
+This command doesn't have any command line options.
+
 == Component Compile
 
 *-f, --values* FILE::
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+* New `commodore cluster list` command ([#135])
 * Vault default values ([#176])
 
 ### Fixed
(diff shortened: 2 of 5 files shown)


srueg:
Nice addition, thanks for the contribution! 👍 

May I suggest to add a `-o/--output` flag to optionally change the output format. This could include the option `json` to enable structured output to further processing, for example:
```
for catalog in $(commodore catalog list | jq -r '.[]'); do
  commodore catalog compile ${catalog};
done
```

srueg (approved):

docs/modules/ROOT/pages/reference/cli.adoc
```diff
@@ -92,6 +92,10 @@ See also xref:local-mode.adoc[local mode documentation].
 *--help*::
   Show catalog clean usage and options then exit.
 
+== Catalog Clean
```
simu, line 95: should be `catalog list`

simu (approved):
LGTM, thanks for the contribution

docs/modules/ROOT/pages/reference/cli.adoc
chrisglass: Ooops! fixed.

PR #183, opened 2020-09-16
More explicit error when failing to add component

chrisglass:
Should "component create" be called outside of a catalog, display some
information in this regard to the user.

The lower layer exception (FileNotFound) is not wrapped in the top-layer
click.Exception() wrapper anymore in order to differentiate exception
types.

## Checklist

- [x] Update tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 * Slightly improved error message in case `commodore component new` is called in
-  a folder that is not a populated (compiled) catalog.
+  a folder that is not a populated (compiled) catalog ([#183])
 * New `commodore cluster list` command ([#135])
 * Vault default values ([#176])
 * Improved `component new` documentation ([#182])
@@ -180,3 +180,4 @@ Initial implementation
 [#176]: https://github.com/projectsyn/commodore/pull/176
 [#177]: https://github.com/projectsyn/commodore/pull/177
 [#182]: https://github.com/projectsyn/commodore/pull/182
+[#183]: https://github.com/projectsyn/commodore/pull/183
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+* Slightly improved error message in case `commodore component new` is called in
+  a folder that is not a populated (compiled) catalog.
 * New `commodore cluster list` command ([#135])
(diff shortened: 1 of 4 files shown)


srueg (changes requested):
Please add a changelog entry.

commodore/helpers.py
```diff
@@ -170,9 +170,6 @@ def relsymlink(srcdir, srcname, destdir, destname=None):
     # https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to
     link_src = os.path.relpath(P(srcdir) / srcname, start=destdir)
     link_dst = P(destdir) / destname
-    try:
-        if link_dst.exists():
-            os.remove(link_dst)
-        os.symlink(link_src, link_dst)
-    except Exception as e:
-        raise click.ClickException(f"While setting up symlinks: {e}") from e
```
simu, line 178: As far as I can tell, the other call sites of `relsymlink` rely on the ClickException being thrown here. Note that the `ClickException` produces an user-friendly error message when it propagates all the way to the top, while other exceptions generate rather unpleasant error messages.

Removing the exception handling here requires that exception handling is introduced at call sites of `relsymlink` (directly or indirectly) to avoid having bare exceptions propagated all the way to the user.

simu (changes requested):
Please make sure that exception handling is introduced to compensate for the removed code in `relsymlink`.

commodore/helpers.py
simu: After discussing with @chrisglass directly, we've decide to let the exceptions bubble up for other call sites and introduce good error messages when such a case pops up.

chrisglass:
> Please add a changelog entry.

Oops! Added.

commodore/helpers.py
chrisglass: Thanks a lot!

simu:
> Please make sure that exception handling is introduced to compensate for the removed code in `relsymlink`.

Follow-up to come up with a concept for and refactor error handling across the whole code base: #184

simu (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+* Slightly improved error message in case `commodore component new` is called in
+  a folder that is not a populated (compiled) catalog.
```
simu, line 13:
```suggestion
  a folder that is not a populated (compiled) catalog ([#183])
```

and add

```
[#183]: https://github.com/projectsyn/commodore/pull/177
```

at the bottom

srueg (approved):

PR #186, opened 2020-09-16
Update component template

simu:
* Fix linting error in `component/app.jsonnet` for components which don't have a `-` in their name.
* Add `--pad-arrays` option to jsonnetfmt arguments in component template Makefile

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 
 * Mypy static type analysis to the tox/CI environments. ([#180])
+* Add `--pad-arrays` to component template jsonnetfmt arguments ([#186])
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


corvus-ch:
As of https://syn.tools/syn/references/style-guide.html#_syntactic_style, we should change the `JSONNETFMT_ARGS` within the `Makefile`. The `--pad-arrays` option is missing. Can you add that to this PR?

simu:
Yes, I will add that argument as well.

It turns out this change is not sufficient, as components which have a `-` in the name actually require the quotes for the key. Will figure out how to do this properly tomorrow.

simu:
Note: currently creating a component whose name starts with a digit completely breaks the template. I would propose that we simply disallow component names starting with a digit.

srueg (approved):

corvus-ch (approved):

PR #187, opened 2020-09-17
Tiny comment addition

chrisglass:
This was said out loud while digging through the code with Simon, the
information as to why we fake argoCD might be interesting to future
travelers.

This is a trivial PR (no functional changes whatsoever)

### DIFF
--- a/commodore/component/compile.py
+++ b/commodore/component/compile.py
@@ -74,6 +74,8 @@ classes:
 """)
 
         # Fake Argo CD lib
+        # We plug "Fake" argoCD configuration here because we don't actually
+        # want to recompile that stuff each when compiling a single component.
         (temp_dir / 'dependencies/lib').mkdir(exist_ok=True)
         with open(temp_dir / 'dependencies/lib/argocd.libjsonnet', 'w') as file:
             file.write("""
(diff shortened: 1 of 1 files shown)


commodore/component/compile.py
```diff
@@ -74,6 +74,8 @@ def compile_component(config: Config, component_path, value_files, search_paths,
 """)
 
         # Fake Argo CD lib
+        # We plug "Fake" argoCD configuration here because we don't actually
+        # want to recompile that stuff each when compiling a single component.
```
srueg, line 78: The reason to fake the library is more that it just wouldn't exist otherwise. You would have to fake it yourself (and provide it with `-J path/to/fake/lib`) because every component relies on this library.

```suggestion
        # We plug "fake" Argo CD library here because every component relies on it
        # and we don't want to provide it every time when compiling a single component.
```

Does that make sense?

srueg (changes requested):

chrisglass:
Github seems to have messed up the suggestion merge... I'll clean it up in another commit.

srueg (approved):

simu (approved):

PR #190, opened 2020-09-22
Manage dependencies using jsonnet-bundler

### LINKED ISSUE #77: Support jsonnet-bundler to manage component jsonnet dependencies
Currently, the mechanisms available in Commodore don't really support pulling in jsonnet dependencies which use [jsonnet-bundler](https://github.com/jsonnet-bundler/jsonnet-bundler).

However, many useful jsonnet libraries (e.g. [kube-prometheus](https://github.com/coreos/kube-prometheus)) make use of jsonnet-bundler. It's nearly impossible to include such a library in a Commodore component at the moment, or pull it in via the `jsonnet_libs` configuration which Commodore supports for 3rd party jsonnet libraries.

### Task deliverables
- [ ] Commodore supports pulling in jsonnet-bundler dependencies at the component level at compile time
- [ ] Commodore supports pulling in jsonnet-bundler dependencies at the global level at compile time

corvus-ch:
Resolves #77 

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,6 +44,7 @@ RUN ./kapitan/inputs/helm/build.sh \
 FROM base AS runtime
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
+      curl \
       git \
       libnss-wrapper \
       openssh-client \
@@ -65,6 +66,10 @@ COPY --from=helm_binding_builder \
 COPY --from=helm_binding_builder \
       /virtualenv/kapitan/dependency_manager/helm/helm_fetch.so \
       /usr/local/lib/python3.8/site-packages/kapitan/dependency_manager/helm/
+
+RUN curl -sLo /usr/local/bin/jb \
+  https://github.com/jsonnet-bundler/jsonnet-bundler/releases/download/v0.4.0/jb-linux-amd64 \
+  && chmod +x /usr/local/bin/jb
 
 COPY ./tools/entrypoint.sh /usr/local/bin/
 
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ using `helm template`.
 * Python 3.6+, with `python3-dev` and `python3-venv` updated
 * [Poetry](https://github.com/python-poetry/poetry)
 * Docker
+* [jsonnet-bundler](https://github.com/jsonnet-bundler/jsonnet-bundler)
 
 ## Getting started
 
@@ -63,6 +64,8 @@ using `helm template`.
       poetry run sh -c '${VIRTUAL_ENV}/lib/python3.*/site-packages/kapitan/inputs/helm/build.sh'
       poetry run sh -c '${VIRTUAL_ENV}/lib/python3.*/site-packages/kapitan/dependency_manager/helm/build.sh'
       ```
+
+    Install jsonnet-bundler according to upstream [documentation](https://github.com/jsonnet-bundler/jsonnet-bundler#install).
 
 1. Setup a `.env` file to configure Commodore (don't use quotes):
 
--- a/commodore/component/compile.py
+++ b/commodore/component/compile.py
@@ -26,7 +26,7 @@ def compile_component(config: Config, component_path, value_files, search_paths,
     value_files = [P(f).resolve() for f in value_files]
     search_paths = [P(d).resolve() for d in search_paths]
     search_paths.append('./dependencies/')
-    search_paths.append(os.path.join(component_path, 'vendor'))
+    search_paths.append(component_path / 'vendor')
     output_path = P(output_path).resolve()
     # Ignore 'component-' prefix in dir name
     component_name = component_path.stem.replace('component-', '')
@@ -92,7 +92,7 @@ local ArgoProject(name) = {};
 
         # Fetch Jsonnet libs
         fetch_jsonnet_libs(config, libs)
-        if os.path.exists(os.path.join(component_path, 'jsonnetfile.json')):
+        if (component_path / 'jsonnetfile.json').exists():
             fetch_jsonnet_libraries(component_path)
 
         # Compile component
--- a/commodore/component/compile.py
+++ b/commodore/component/compile.py
@@ -6,7 +6,7 @@ import tempfile
 import click
 
 from commodore.config import Config, Component
-from commodore.dependency_mgmt import fetch_jsonnet_libs
+from commodore.dependency_mgmt import fetch_jsonnet_libs, fetch_jsonnet_libraries
 from commodore.helpers import kapitan_compile, relsymlink
 from commodore.postprocess import postprocess_components
 from git import Repo
@@ -25,6 +25,8 @@ def compile_component(config: Config, component_path, value_files, search_paths,
     component_path = P(component_path).resolve()
     value_files = [P(f).resolve() for f in value_files]
     search_paths = [P(d).resolve() for d in search_paths]
+    search_paths.append('./dependencies/')
+    search_paths.append(os.path.join(component_path, 'vendor'))
     output_path = P(output_path).resolve()
     # Ignore 'component-' prefix in dir name
     component_name = component_path.stem.replace('component-', '')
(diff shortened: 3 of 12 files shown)


srueg (approved):

srueg (changes requested):
Now the component compilation fails because `dependencies/` is not in the search path anymore...

README.md
```diff
@@ -64,6 +65,9 @@ using `helm template`.
       poetry run sh -c '${VIRTUAL_ENV}/lib/python3.*/site-packages/kapitan/dependency_manager/helm/build.sh'
       ```
 
+    Install jsonnet-bundler according to upstream [documentation](https://github.com/jsonnet-bundler/jsonnet-bundler#install).
+    For OS X, jsonnet-bundler is available as [Homebrew](https://brew.sh/) package.
```
tobru, line 69: I'd prefer to leave this line out, otherwise it gives the feeling that it's OS X biased or we would need to add the info for other OS'es as well. I'd leave how to install it up to the user.

docs/modules/ROOT/pages/index.adoc
```diff
@@ -18,6 +18,14 @@ templates. Commodore sets up an https://kapitan.dev/inventory/[inventory] by
 downloading all required components and symlinking the relevant classes into
 the inventory structure which Kapitan understands.
 
+Components can define dependencies to Jsonnet libraries using https://github.com/jsonnet-bundler/jsonnet-bundler[jsonnet-bundler].
```
tobru, line 21: This feels a bit scarce. Shouldn't we add a bit more documentation (example, how-to, ...) about this feature, as it feels to be an important feature for the project.

tobru (changes requested):
Other than the inline comments: Should Jsonnet Bundlet not added to the Dockerfile as well?

commodore/dependency_mgmt.py
```diff
@@ -195,3 +197,39 @@ def fetch_jsonnet_libs(config, libs):
...
+    jsonnetfile = {
+        "version": 1,
+        "dependencies": dependencies,
+        "legacyImports": True,
+    }
+
+    with open("jsonnetfile.json", "w") as file:
+        file.write(json.dumps(jsonnetfile, indent=4))
+
+    if call(['jb', 'install']) != 0:
```
simu, line 234: Probably want to produce a useful error here if `jb` cannot be found

simu (changes requested):
Please fix the test failures, looks like component compile breaks with these changes

corvus-ch:
> Now the component compilation fails because dependencies/ is not in the search path anymore...
> We either add it to the search path for component compilation only or run jb also there.

I fixed this by splitting the creation of the `jsonnetfile.json` and the call to `jb`. I also removed `vendor` from the default search path and only add it where needed.

For the `commodore component compile`, `jb` is called for only that component. Search paths are adapted accordingly.

Dockerfile
```diff
@@ -41,6 +41,10 @@ COPY --from=builder /usr/local/lib/python3.8/site-packages/kapitan ./kapitan
 RUN ./kapitan/inputs/helm/build.sh \
  && ./kapitan/dependency_manager/helm/build.sh
 
+FROM golang:1.15 AS jsonnet-bundler
+
+RUN go get github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb
```
srueg, line 46: Let's download a specific version from the [releases](https://github.com/jsonnet-bundler/jsonnet-bundler/releases).

commodore/component/compile.py
```diff
@@ -90,6 +92,8 @@ def compile_component(config: Config, component_path, value_files, search_paths,
 
         # Fetch Jsonnet libs
         fetch_jsonnet_libs(config, libs)
+        if os.path.exists(os.path.join(component_path, 'jsonnetfile.json')):
```
srueg, line 95: Since `component_path` is already a path type, it should be possible to do
```suggestion
        if (component_path / 'jsonnetfile.json').exists():
```

commodore/component/compile.py
```diff
@@ -25,6 +25,8 @@ def compile_component(config: Config, component_path, value_files, search_paths,
     component_path = P(component_path).resolve()
     value_files = [P(f).resolve() for f in value_files]
     search_paths = [P(d).resolve() for d in search_paths]
+    search_paths.append('./dependencies/')
+    search_paths.append(os.path.join(component_path, 'vendor'))
```
srueg, line 29:
```suggestion
    search_paths.append(component_path / 'vendor')
```

srueg (changes requested):

Dockerfile
```diff
@@ -44,6 +44,7 @@ RUN ./kapitan/inputs/helm/build.sh \
 FROM base AS runtime
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
+      curl \
```
simu, line 47: That's gonna improve the debugging experience when running Commodore in Docker quite a bit :+1:

docs/modules/ROOT/pages/dependencies.adoc
```diff
@@ -0,0 +1,46 @@
+= Manage dependencies
+
+Commodore components can declare dependencies to Jsonnet libraries.
+Those dependencies are resolved and downloaded using https://github.com/jsonnet-bundler/jsonnet-bundler[jsonnet-bundler].
+
+Define dependencies in a `jsonnetfile.json` file.
+This file must be placed at the root of the components file structure.
```
simu, line 7:
```suggestion
This file must be placed at the root of the component's file structure.
```

docs/modules/ROOT/pages/dependencies.adoc
```diff
@@ -0,0 +1,46 @@
...
+====
+Jsonnet-bundler uses the file `jsonnetfile.lock.json` to freeze dependency versions for a project.
+Lock files of libraries aren't taken into consideration and within this setup, components are libraries.
+If your component require a specific version, you have to declare this in the `version` field within `jsonnetfile.json`.
+====
+
+[TIP]
+====
+When compiling components with `commodore component compile`, jsonnet-bundler will create the file `jsonnetfile.lock.json` and the folder `vendor`.
+Add them to your `.gitignore`.
```
simu, line 45: Probably worth updating the `.gitignore` in the component template? Doesn't hurt if the entries are there but not used.

commodore/component/compile.py
simu: looks like the suggestion was not fully applied

commodore/component/compile.py
simu: looks like the suggestion was not fully applied

simu (approved):
LGTM, apart from some not yet applied suggestions from the last round of code review

commodore/component/compile.py
corvus-ch:
```suggestion
        if (component_path / 'jsonnetfile.json').exists():
```

corvus-ch:
>  […] apart from some not yet applied suggestions from the last round of code review

With greetings from force push. I like that GitHub reopened those.

docs/modules/ROOT/pages/dependencies.adoc
```diff
@@ -0,0 +1,46 @@
...
+====
+If two components depend on the same Jsonnet library, they must require the same version.
+See https://github.com/jsonnet-bundler/jsonnet-bundler#current-limitations[jsonnet-bundler - Current Limitations]
+====
+
+[IMPORTANT]
+====
+Jsonnet-bundler uses the file `jsonnetfile.lock.json` to freeze dependency versions for a project.
+Lock files of libraries aren't taken into consideration and within this setup, components are libraries.
+If your component require a specific version, you have to declare this in the `version` field within `jsonnetfile.json`.
```
srueg, line 39:
```suggestion
If your component requires a specific version, you have to declare this in the `version` field within `jsonnetfile.json`.
```

srueg (approved):

simu:
> > […] apart from some not yet applied suggestions from the last round of code review
> 
> With greetings from force push. I like that GitHub reopened those.

I actually reopened them manually rather than write a new identical suggestion. GitHub isn't quite that smart yet unfortunately.

tobru (approved):

PR #191, opened 2020-09-24
Subprocess changes and dedent

chrisglass:
This branch provides two nitpicky non-functional changes:

1. Switch from using os.system() to the higher-level subprocess.call() (the recommended interface for shelling out in python). In particular this allows to retrieve non-zero exit codes instead of getting the syscall + exit code integer from os.system.
2. Use textwrap.dedent() to have the yaml snippets aligh with the rest of the code instead of "leaking" to the left-hand side of the text file.

These changes are entirely "cosmetic" and should not affect the functionality at all.

## Checklist

### DIFF
--- a/commodore/component/compile.py
+++ b/commodore/component/compile.py
@@ -84,8 +84,8 @@ def compile_component(config: Config, component_path, value_files, search_paths,
                 local ArgoProject(name) = {};
 
                 {
-                App: ArgoApp,
-                Project: ArgoProject,
+                  App: ArgoApp,
+                  Project: ArgoProject,
                 }"""))
 
         # Fetch Jsonnet libs
--- a/commodore/component/compile.py
+++ b/commodore/component/compile.py
@@ -2,6 +2,7 @@ import os
 from pathlib import Path as P
 import shutil
 import tempfile
+from textwrap import dedent
 
 import click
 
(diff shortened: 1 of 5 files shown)


simu (approved):
LGTM

commodore/component/compile.py
```diff
@@ -42,51 +43,48 @@ def compile_component(config: Config, component_path, value_files, search_paths,
...
-  App: ArgoApp,
-  Project: ArgoProject,
-}
-""")
+            file.write(dedent("""
+                local ArgoApp(component, namespace, project='', secrets=true) = {};
+                local ArgoProject(name) = {};
+
+                {
+                App: ArgoApp,
```
srueg, line 85:
```suggestion
                  App: ArgoApp,
```

commodore/component/compile.py
```diff
@@ -42,51 +43,48 @@ def compile_component(config: Config, component_path, value_files, search_paths,
...
+                Project: ArgoProject,
```
srueg, line 86:
```suggestion
                  Project: ArgoProject,
```

tests/test_component_compile.py
```diff
@@ -14,15 +18,14 @@ def _prepare_component(tmp_path, component_name='test-component'):
...
-{
-  "test_service_account": kube.ServiceAccount('test') {
-    metadata+: {
-      namespace: params.namespace,
-    },
-  },
-}
-''')
+        file.write(dedent('''
+            {
```
srueg, line 22: Indentation broke a little here.

srueg (changes requested):

commodore/component/compile.py
chrisglass: Oh, good catch! It's valid JSONnet, so it did escape the unit tests. Will change (and other places too of course).

tests/test_component_compile.py
chrisglass: Yep, same problem as above. I'll fix it.

Good catch!

chrisglass: Done.

srueg (approved):

PR #195, opened 2020-09-28
Make inventory hierarchy dynamic

### LINKED ISSUE #93: Make the Inventory Structure Configurable
# Background
Currently the inventory structure is hard coded into Commodore:
* `global.common`
* `global.cloud.<cloud-provider>`
* `global.cloud.<cloud-provider>.<cloud-region>`
* `global.distribution.<k8s-distribution>`
* `<customer>.<cluster>`

This limits the flexibility of different setups and makes changes to this hierarchy more difficult.

Similar to Puppet Hiera, this structure should be configurable outside of Commodore.

# Proposal
Implement a config file which can define the inventory structure. This config file should be created in the global config Git repo with a specific name (e.g. `syn.yml`). It should allow to use facts as placeholders for names in the structure. The leaf of an entry is always a class (i.e. `.yml` file).

The referenced classes should be optional and if they don't exist they should be skipped. I.e. if a cloud doesn't have a region, no `cloud/${cloud}/${region}.yml` file needs to exist.

Possible example:
```yaml
hierarchy:
  - global/common
  - global/cloud/${cloud}
  - global/cloud/${cloud}/${region}
  - global/distribution/${distribution}
  - global/lieutenant-instance/${lieutenant-instance}
  - ${tenant}/common
  - ${tenant}/${cluster}
```
# References
* Puppet Hiera config: https://puppet.com/docs/hiera/3.3/configuring.html

corvus-ch:
Resolves #93 

## Checklist

- [x] Update tests.


corvus-ch:
This is currently being blocked by https://github.com/deepmind/kapitan/pull/610.

corvus-ch:
I am currently writing the documentation. @simu @srueg could you please already review the code changes? Please note the two commits starting with `TEMP` in the message. Those two commits will be removed before this PR can be merged. They are workarounds for changes to be done elsewhere.

commodore/cluster.py
```diff
@@ -19,89 +23,105 @@ def fetch_cluster(cfg, clusterid):
...
+            'global.cloud.${facts:cloud}',
+            'global.cloud.${facts:cloud}.${facts:region}',
+            '${cluster:tenant}.${cluster:name}',
+
+        ]}, common)
+
+
+def render_params(cluster, target: str):
+    facts = cluster['facts']
+    for fact in ['distribution', 'cloud']:
```
srueg, line 87: Is this still something we should enforce?
I'm just thinking how generic these really are. We don't have any dependency on hard coded facts anymore anywhere else...

commodore/cluster.py
```diff
@@ -19,89 +23,105 @@ def fetch_cluster(cfg, clusterid):
...
 
+    # TODO Remove after deprecation phase.
+    if 'region' in facts:
+        data['parameters']['cloud']['region'] = facts['region']
+
+    return data
+
+
+def params_file(target: str):
+    return P('inventory/classes/params') / f"{target}.yml"
```
srueg, line 120:
```suggestion
    return P('inventory', 'classes', 'params') / f"{target}.yml"
```

commodore/compile.py
```diff
@@ -59,41 +64,36 @@ def _fetch_customer_config(cfg, customer_id):
...
-    if not target_yml.is_file():
-        raise click.ClickException(f"Invalid target: {target_name}")
-    click.echo(f" > Using target: {target_name}")
+    if not target_file(target).is_file():
+        raise click.ClickException(f"Invalid target: {target}")
+    click.echo(f" > Using target: {target}")
 
     click.echo(' > Reconstructing Cluster API data from target')
-    cluster = reconstruct_api_response(target_yml)
+    cluster = reconstruct_api_response(target)
```
srueg, line 96: Since it looks like we only need the cluster and tenant IDs anymore, can we change `reconstruct_api_response()` to only get these two values?

srueg (changes requested):

commodore/cluster.py
```diff
@@ -19,89 +23,105 @@ def fetch_cluster(cfg, clusterid):
...
+
+    return response
+
+
+def render_target(target: str, components: Iterable[str]):
+    classes = [f"params.{target}"]
+
+    for component in components:
+        defaults_file = P('inventory/classes/defaults') / f"{component}.yml"
+        if defaults_file.is_file():
```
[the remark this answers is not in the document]
corvus-ch: I am not sure. Maybe @simu does have an opinion on that.

commodore/cluster.py
corvus-ch: You are right. This is no longer needed. I will remove it.

corvus-ch: Um wait! We need to keep it until we remove the deprecated `cluster.dist` and `cloud.provider` parameters.

commodore/compile.py
corvus-ch: Refactored to `read_cluster_and_tenant(target: str) -> Tuple[str, str]`.

pyproject.toml
```diff
@@ -13,7 +13,7 @@ include = [
 [tool.poetry.dependencies]
 python = "^3.6"
 # Require kapitan >= 0.27 to help poetry to resolve dependencies
-kapitan = ">=0.29"
+kapitan = { git = "https://github.com/deepmind/kapitan.git", branch = "bfeda63836e5fbd220f3e4cd4bd490d9e1a7eb09" }
```
srueg, line 16:
```suggestion
kapitan = { git = "https://github.com/deepmind/kapitan.git", rev = "bfeda63836e5fbd220f3e4cd4bd490d9e1a7eb09" }
```

corvus-ch:
Turns out deepmind/kapitan#610 was not needed after all. I removed the Kapitan update as it was braking the build.

commodore/cluster.py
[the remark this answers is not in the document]
tobru: I would say: yes.

docs/modules/ROOT/pages/reference/hierarchy.adoc
```diff
@@ -0,0 +1,89 @@
+= Configure the inventory hierarchy
```
tobru, line 1:
```suggestion
= Configure the Inventory Hierarchy
```

docs/modules/ROOT/pages/reference/hierarchy.adoc
```diff
@@ -0,0 +1,89 @@
...
+
+The https://kapitan.dev/inventory/[Kapitan inventory] used by Commodore builds on top of https://reclass.pantsfullofunix.net/[reclass].
```
tobru, line 3: Better link to the fork? https://github.com/kapicorp/reclass

docs/modules/ROOT/pages/reference/concepts.adoc
```diff
@@ -11,6 +11,12 @@ low to high priority):
 
 * Component parameter default values
 * Global generic default values
+
+That global generic default allows then to include further classes and build up any hierarchy.
+This is enabled by the ability to use cluster facts when including classes.
+
+The hierarchy could then extended by the following levels.
```
srueg, line 18:
```suggestion
The hierarchy could then be extended by the following levels.
```

docs/modules/ROOT/pages/reference/hierarchy.adoc
```diff
@@ -0,0 +1,89 @@
...
+<3> Clone of the git repository holding your global configuration hierarchy.
+<4> Included by the clusters target and thus the starting point for your inventory hierarchy.
+<5> Holds parameters set by commodore along with cluster facts retrieved from Lieutenant.
+    Included by the clusters target.
+    This file is generated by Commodore.
+<6> Clone of the git repository holding the configuration hierarchy of the clusters tenant.
+<7> Kapitan target for the cluster.
+    This file is generated by commodore.
+
+To build up your configuration hierarchy, included classes in `common.yml` within your global configuration inventory.
```
srueg, line 37:
```suggestion
To build up your configuration hierarchy, include classes in `common.yml` within your global configuration inventory.
```

docs/modules/ROOT/pages/reference/hierarchy.adoc
```diff
@@ -0,0 +1,89 @@
...
+----
+classes:
+  - global.distribution.${facts:distribution}
+  - global.cloud.${facts:cloud}
+  - global.cloud.${facts:cloud}.${facts:region}
+  - ${cluster:tenant}.${cluster:name}
+----
+
+Reclass requires parameters to be defined when used in interpolation.
+When building a hierarchy that have optional parts, one must be creative.
```
srueg, line 65:
```suggestion
When building a hierarchy that has optional parts, one must be creative.
```

tobru (approved):
Two small nitpicks - otherwise LGTM. Let's see how we continue with the documentation, I think the whole hierarchy stuff needs more. But for the time being that's good enough.

docs/modules/ROOT/partials/nav-reference.adoc
```diff
@@ -2,3 +2,4 @@
 * xref:commodore:ROOT:reference/architecture.adoc[Architecture]
 * xref:commodore:ROOT:reference/commands.adoc[Commands]
 * xref:commodore:ROOT:reference/cli.adoc[Command Line Options]
+* xref:commodore:ROOT:reference/hierarchy.adoc[Inventory hierarchy]
```
srueg, line 5:
```suggestion
* xref:commodore:ROOT:reference/hierarchy.adoc[Inventory Hierarchy]
```

srueg (changes requested):

commodore/cluster.py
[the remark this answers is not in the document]
simu: Commodore used to print a warning about 'old-style components' when a component did not have a `defaults.yml`, but it appears that has been removed at some point.

docs/modules/ROOT/pages/reference/concepts.adoc
```diff
@@ -11,6 +11,12 @@ low to high priority):
 
 * Component parameter default values
 * Global generic default values
+
+That global generic default allows then to include further classes and build up any hierarchy.
```
simu, line 15: With the change the whole structure of this section is pretty confusing, as the configuration value sources can now arbitrarily be changed.

I think the part of the section which starts with "Commodore defines the following hierarchy" needs to be completely replaced with a description of the new dynamic inventory mechanism, and a description of how component defaults fit into the new picture.

docs/modules/ROOT/pages/reference/hierarchy.adoc
simu: The Microsoft style guide (which we're using according to our documentation) says to use sentence case for titles (cf. https://docs.microsoft.com/en-us/style-guide/capitalization#sentence-style-capitalization-in-titles-and-headings)

docs/modules/ROOT/pages/reference/hierarchy.adoc
```diff
@@ -0,0 +1,89 @@
+= Configure the inventory hierarchy
+
+The https://kapitan.dev/inventory/[Kapitan inventory] used by Commodore builds on top of https://reclass.pantsfullofunix.net/[reclass].
+Reclass allows to do interpolation of parameters.
```
simu, line 4: Note: Upstream Reclass doesn't allow parameters in classes. That's one of the features of Kapitan's fork: https://github.com/kapicorp/reclass/blob/develop/README-extensions.rst#use-references-in-class-names

simu (commented):
LGTM overall. Some comments regarding the docs.

docs/modules/ROOT/pages/reference/hierarchy.adoc
srueg: Note: It's not really the Kapicorp fork, but the salt-formula fork: https://github.com/salt-formulas/reclass
(which in turn is a fork of the original madduck reclass: https://github.com/madduck/reclass)

docs/modules/ROOT/pages/reference/hierarchy.adoc
corvus-ch: The vale checker does not assert this but the Microsoft style guide asks for sentence case in titles.

commodore/cluster.py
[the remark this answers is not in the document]
corvus-ch: This is already covered. See https://github.com/projectsyn/commodore/blob/159b294627313cf54a237e43d835e493644dcd8b/commodore/dependency_mgmt.py#L30 and https://github.com/projectsyn/commodore/blob/159b294627313cf54a237e43d835e493644dcd8b/commodore/component/compile.py#L129.

docs/modules/ROOT/pages/reference/concepts.adoc
corvus-ch: I have rewritten that part.

docs/modules/ROOT/pages/reference/hierarchy.adoc
corvus-ch: I have now explicitly stated, that a fork is being used. Including link to the extension documentation of that fork.

simu (approved):

srueg (approved):

commodore/cluster.py
```diff
@@ -67,6 +69,18 @@ def update_target(cfg: Config, target):
...
+    # TODO Remove before merge!
+    common = P('inventory/classes/common.yml')
+    os.makedirs(common.parent, exist_ok=True)
+    yaml_dump({
+        'classes': [
+            'global.distribution.${facts:distribution}',
+            'global.cloud.${facts:cloud}',
+            'global.cloud.${facts:cloud}.${facts:region}',
+            '${cluster:tenant}.${cluster:name}',
+
```
srueg, line 75:
```suggestion
            'global.lieutenant-instance.${facts:lieutenant-instance}',
```

srueg: For some reason the `lieutenant-instance` class is not properly included if I try to compile the `c-old-fire-5788` cluster.

commodore/cluster.py
```diff
@@ -50,6 +50,8 @@ def render_target(target: str, components: Iterable[str]):
         if defaults_file.is_file():
             classes.append(f"defaults.{component}")
 
+    # TODO Remove before merge!
+    classes.append('common')
```
srueg, line 48: This needs to be included after `global.common`

commodore/cluster.py
srueg: Hmm, seems like the referencing breaks because the order of import changed: As mentioned above, `global.common` needs to be included before `common` for it to work properly.

srueg: I did some tests and I fear we have to split the hierarchy definition and parameters into two files.
Currently both is defined in [common.yml](https://git.vshn.net/syn/commodore-defaults/-/blob/dynamic_hierarchy/common.yml). But this has the effect, that references used in there (e.g. `VAULT_ADDR: ${secret_management:vault_addr}`) are not properly overwritten, because they are evaluated first, before including other classes.

We could for example leave only the class imports in `common.yml` and move all the `parameters` definitions into a new file (e.g. `global.params`) which needs to be included as the first class.

srueg (changes requested):

commodore/cluster.py
corvus-ch: Resolved by splitting class includes and parameters into different files.

simu (approved):

srueg (approved):

PR #199, opened 2020-10-01
Do not use `pip` to install Poetry

corvus-ch:
Having Poetry installed within the same scope as the application, can
result in unexpected behaviour. Using their install script, puts poetry
in isolation and no interference should occur.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,18 @@
 FROM docker.io/python:3.8.6-slim-buster AS base
 
 ENV HOME=/app
-ENV PATH=${PATH}:${HOME}/.poetry/bin
 
 WORKDIR ${HOME}
 
 FROM base AS builder
 
+ENV PATH=${PATH}:${HOME}/.poetry/bin
+
 RUN apt-get update && apt-get install -y --no-install-recommends \
       build-essential \
       curl \
  && rm -rf /var/lib/apt/lists/* \
- && curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python - \
+ && curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python - --version 1.1.0 \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,15 +1,17 @@
...
 WORKDIR ${HOME}
 
 FROM base AS builder
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
       build-essential \
+      curl \
  && rm -rf /var/lib/apt/lists/* \
- && pip install poetry \
+ && curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python - \
```
srueg, line 14:
```suggestion
 && curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python - --version 1.1.0 \
```

Dockerfile
```diff
@@ -1,15 +1,17 @@
 FROM docker.io/python:3.8.6-slim-buster AS base
 
 ENV HOME=/app
+ENV PATH=${PATH}:${HOME}/.poetry/bin
```
srueg, line 4: You can do this in a single `ENV` directive (adding a backslash):
```suggestion
    PATH="${PATH}:${HOME}/.poetry/bin"
```

srueg (changes requested):

Dockerfile
```diff
@@ -1,15 +1,17 @@
 FROM docker.io/python:3.8.6-slim-buster AS base
 
 ENV HOME=/app
```
simu, line 3:
```suggestion
ENV HOME=/app \
```

simu (approved):

Dockerfile
srueg: My bad, sorry. It would need to be `PATH="${PATH}:/app/.poetry/bin"` since `$HOME` is not set at that point.

srueg (approved):

PR #205, opened 2020-10-05
Black

chrisglass:
This branch introduces the [black](https://github.com/psf/black) formatter to the project. black is a strongly-opinionated formatter similar in usage to gofmt in the golang world.

This branch is best reviewed per-commit, since it introduces a lot of changes to all files:
- 4e3d204  only deals with adding CI wiring in place to make sure formatting is enforced during PR reviews.
- 2c623df Touches up the README by adding a small note about black formatting being enforced by CI
- 61373f3 is the main commit that only contains mechanical changes produced by running the formatter on the source tree.

It seems to me like this PR should be open as "fast" as possible for awareness. This branch is quite easy to rebase, so I'm happy to do that should other branches get merged in the mean time, just let me know on this PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,6 +14,7 @@ jobs:
           - lint_pylint
           - lint_bandit
           - lint_mypy
+          - lint_black
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
--- a/README.md
+++ b/README.md
@@ -111,6 +111,9 @@ using `helm template`.
    - Write a line of application code, make the test pass
    - Repeat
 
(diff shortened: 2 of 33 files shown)


simu (approved):

srueg (approved):
We should probably add black as a dev dependency to pyproject.toml?

chrisglass:
> We should probably add black as a dev dependency to pyproject.toml?

I don't know - do you think so? I tend not to use poetry because I'm more used to the "traditional" tools, but I'm happy to add an entry there in this PR if it helps your (or someone else's) workflow.

corvus-ch (approved):
We do not need to add black as a dev depencendy. Tox takes care of that.

PR #222, opened 2020-10-19
Inventory postprocessing filters

### LINKED ISSUE #156: Refactor postprocessing to define filters in inventory
## Context

Currently, we're defining postprocessing filters in each component in a dedicated file `postprocess/filters.yml`. This has some upsides -- we know which component defines each filter, we know where the component filters are based on the definition file -- but also some downsides -- we don't get reclass features, e.g. variable references for free, we have some configuration that's not part of the inventory.

## Alternatives

The current implementation is an alternative for the proposed change

## Proposal

Refactor the Commodore postprocessing to extract filters to run from the inventory, e.g. key `parameters.commodore.postprocess.filters`. This key could be extended by component classes that want to define postprocessing filters, the same way each component provides entries in `parameters.kapitan.compile`.

This change would allow us to extract filters from the parsed inventory, with all reclass references already resolved.

We would need to adjust some of the code which executes the postprocessing filters, and probably require some more information to be provided in the filter definitions, especially the path to the filter code for custom jsonnet filters would have to be given in each filter definition.

The idea would be to introduce "inventory filter definitions" for new components and gradually migrate existing components to the new method, before fully removing support for `postprocess/filters.yml`.

## References

This idea first popped up in #155 where we considered options to avoid having to implement custom code to resolve reclass references "by hand" in Commodore.

simu:
Resolves #156

## Checklist

- [x] Update tests.


chrisglass (approved):
Looks good!

srueg (approved):

Issue #223, opened 2020-10-20
Define a Kapitan target per component


simu:
## Context

We've found some instances (e.g. #156) where it would be beneficial to have a Kapitan target per component. For #156, having specific targets per component would allow us to reduce the amount of information that must be provided for the postprocessing filter definition (e.g. the component defining the filter would be clear based on the target).

There are some challenges in the implementation, if we want to make this change in a backwards-compatible manner. The current architecture relies on the single target Kapitan inventory for a number of features, the big ones being the dynamic hierarchy and component discovery. Additionally, some approaches for implementing this change would require refactoring all component classes.

Below we outline an approach which should be backwards compatible for the dynamic hierarchy and component discovery implementation, while not requiring any refactoring in the existing components.

## Proposal

The best approach which doesn't require component refactoring and allows us to reuse the current implementation for the dynamic hierarchy and component discovery is to reorganize the Kapitan inventory slightly.

The new inventory structure will be constructed with empty files in `classes/components/` for each component. This way, the existing class includes of form `components.<component-name>` do not pull in the component class which has the `parameters.kapitan` key anymore. The actual component classes will be made available in `classes/_components` (exact naming TBD). 

With this inventory structure, each target can have the dynamic hierarchy class include of `global.commodore` without defining compilation instructions for each component. Each target still needs to include all component default classes, to ensure all inventory references can be resolved. Additionally each target will need to include class `_components.<component-name>` to pull in the Kapitan compile instructions for the targeted component.

A target will have the following form:

```
classes:
  - [ ... all component defaults for the cluster ... ]
  - global.commodore
  - _components.<comonent-name>

parameters:
  kapitan:
    vars:
      target: <component-name>
```

## Alternatives

An alternative approach which requires refactoring of all components would be to change components to define their Kapitan configuration in `parameters.<component-name>.kapitan` instead of `parameters.kapitan`. This allows us to continue using the component classes in the hierarchy exactly the same way as previously.

Commodore could then generate individual targets per component where `parameters.kapitan` is extended as follows

```
parameters:
  kapitan: ${<component_name>:kapitan}
```

Each component is then responsible that their `parameters.<component-name>.kapitan` entry has `vars.target` set to `<component-name>`.

srueg:
Reclass also has the concept of [aplications](https://reclass.pantsfullofunix.net/concepts.html) (basically a list of strings?). Maybe we can use this to specify the components to include in the hierarchy? (To prevent creating these empty class files).
[...]

PR #225, opened 2020-10-20
Create a Kapitan target for each component

simu:
WIP PR for a PoC for #223 

Notable changes:
* Components are included in the hierarchy via `applications` instead of `classes` (64b82ee8fe6d03f9176bbbfd14670ef1291ef94f). Component includes in the commodore-defaults and tenant repos need to be moved to `applications`, as described in the commit message.
* The refactored code implicitly disallows a component with name "cluster" to exist, as we're using the target name "cluster" for the bootstrapping target, so we can actually discover components, find component versions, generate Kapitan refs, etc. from a well-known target.
* The refactored code does not support compiling a component in a Kapitan target whose name is not the component name.
* Due to the previous point, the single-component compilation had to be adjusted to create a target named the same as the component. Therefore, component compilation would overwrite the intermediate result of catalog compilation for the component, if it exists. To avoid this, single-component compilation appends "standalone" to the requested output dir (TBD if we can do this in a nicer way).

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/compile.py
+++ b/commodore/compile.py
@@ -31,9 +31,6 @@ from .postprocess import postprocess_components
 from .refs import update_refs
 
 
-TARGET = "cluster"
-
-
 def _fetch_global_config(cfg, cluster):
     config = cluster["base_config"]
     click.secho("Updating global config...", bold=True)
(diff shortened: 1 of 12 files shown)


commodore/compile.py
```diff
@@ -125,25 +125,26 @@ def _local_setup(config, cluster_id, target):
...
     # versions are assumed to be defined in the inventory key
     # 'parameters.component_versions'
     reset_reclass_cache()
-    kapitan_inventory = inventory_reclass("inventory")["nodes"][TARGET]
-    versions = kapitan_inventory["parameters"].get("component_versions", None)
+    cluster_inventory = inventory_reclass("inventory")["nodes"]["cluster"]
+    versions = cluster_inventory["parameters"].get("component_versions", None)
     if versions and not config.local:
         set_component_overrides(config, versions)
-        update_target(config, TARGET)
```
simu, line 141: I believe this call to `update_target` has been a noop for a while now, but I'm not completely convinced yet.

Issue #198, opened 2020-10-01
Configurable revisions for global and tenant config repositories


corvus-ch:
Commodore always clones the master branch of the global and the tenant config repositories. Using the `--local` flag, one can change this behavior. While hacking around with Commodore, using local mode might not be possible. Having the ability to use other references (tag, commit) would solve this.

This would have helped in #195. There we had to add a temporary commit that had to be removed right before the merge.

corvus-ch:
This can be implemented simply by implementing some flags with commodore. This most certainly would be a 80/20 solution. The drawback would be, the version pinning is not persistent.

In order to tackle this, we would need to extend the Tenant and Cluster objects. Both objects would allow to set version of the global and the tenant config repository. Tenant would take precedence over the global default and Cluster would take precedence over Tenant.

Which approach shall we follow?

simu:
I think long term, we'll want the solution which allows specifying a version of the global and tenant repository in the Tenant and Cluster API objects.

Please note that this issue is loosely coupled to #171, which aims to make the global configuration repository configurable per tenant/cluster, rather than requiring the counterintuitive GLOBAL_GIT_BASE parameter.

Short term, both features could be implemented by (ab)using the existing facilities to define custom facts per cluster (tenant-scoped custom facts are not supported at the moment).

Issue #171, opened 2020-09-07
Location of global-common Git repo configurable


tobru:
## Context

Make the global common config Git repo location configurable via Lieutenant API. Get rid of COMMODORE_GLOBAL_GIT_BASE. Goes hand in hand with #170

corvus-ch:
With #198 done, `COMMODORE_GLOBAL_GIT_BASE` is now only used as a fallback (maintain compatibility to older versions Lieutenant API). It is still in use to auto discover components not explicitly listed in `commodore.yml` within the global config git repo.

corvus-ch:
Having this discussed with @simu, we concluded to just drop auto discovery for now. As a consequence, each and every component must be listed in `commodore.yml` within the global config git repo.

corvus-ch:
I have extracted the drop part to #229. @tobru do you agree to close this one.

PR #226, opened 2020-10-21
Get git repository info from API

corvus-ch:
Implementation for projectsyn/commodore#198 and projectsyn/commodore#171.

## Checklist

- [x] Update tests.


CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * Clear Jsonnet lock file ([#215])
 * Also make arrays mergeable with `makeMergeable` helper ([#217])
 
+### Changed
+- Git repository urls are now retrieved from the Lieutenant API ([#226])
```
simu, line 19:
```suggestion
- Git repository URLs are now retrieved from the Lieutenant API ([#226])
```

commodore/cluster.py
```diff
@@ -13,14 +13,79 @@
...
+                % self.__cluster["id"]
+            )
+        return repo_url
+
+    def tenant(self) -> str:
+        return self.__tenant["id"]
+
+    def facts(self) -> Dict[str, str]:
+        if "facts" not in self.__cluster:
+            return {}
```
simu, line 75: Is it allowed that a cluster API response doesn't have the field facts? Or should this be an error?

commodore/cluster.py
```diff
@@ -13,14 +13,79 @@
...
+        self.__cfg = cfg
+        self.__cluster = cluster_response
+        self.__tenant = tenant_response
+        if (
+            "tenant" not in self.__cluster
+            or self.__cluster["tenant"] != self.__tenant["id"]
+        ):
+            raise click.ClickException("Customer id mismatch")
+
+    def id(self) -> str:
```
simu, line 35:
```suggestion
    @property
    def id(self) -> str:
```

By decorating getters with `@property` they can be used as `cluster.id` instead of `cluster.id()`.

commodore/compile.py
```diff
@@ -25,57 +26,57 @@
...
-    config = cluster["base_config"]
+def fetch_global_config(cfg, cluster: Cluster):
     click.secho("Updating global config...", bold=True)
     repo = git.clone_repository(
-        f"{cfg.global_git_base}/{config}.git", "inventory/classes/global", cfg
+        cluster.global_git_repo_url(), "inventory/classes/global", cfg
     )
+    rev = cluster.global_git_repo_revision()
+    if rev != HEAD:
+        git.checkout_version(repo, rev)
```
simu, line 45: Consider getting rid of the constant `HEAD` and just return `None` when there's no revision overrides.

Then this could be simplified to

```suggestion
    if rev:
        git.checkout_version(repo, rev)
```

docs/modules/ROOT/pages/reference/cli.adoc
```diff
@@ -33,19 +33,14 @@ This command doesn't have any command line options.
...
-field `base_config`.
+Commodore uses the global Git base URL when doing component autodiscovery.
+Components are searched in `GLOBAL_GIT_BASE/commodore-components`, unless specified otherwise in the configuration, see xref:reference/architecture.adoc#_component_discovery[Component discovery].
 +
-By default, Commodore also uses the global Git base URL when doing component
-autodiscovery. Components are searched in
-`GLOBAL_GIT_BASE/commodore-components`, unless specified otherwise in the
-configuration, see
-xref:reference/architecture.adoc#_component_discovery[Component discovery].
+In case, the Lieutenant API doesn't return the global Git repository URL, the global git base will also be used to determine URL of that repository.
```
simu, line 41:
```suggestion
If the Lieutenant API doesn't return the global Git repository URL, the global Git base will also be used to determine the URL of that repository.
```

docs/modules/ROOT/pages/reference/cli.adoc
```diff
@@ -33,19 +33,14 @@ This command doesn't have any command line options.
...
+It URL will be discovered as `GLOBAL_GIT_BASE/commodore-defaults.git`.
```
simu, line 42:
```suggestion
The fallback global repository URL is `GLOBAL_GIT_BASE/commodore-defaults.git`.
```

docs/modules/ROOT/pages/reference/cli.adoc
```diff
@@ -33,19 +33,14 @@ This command doesn't have any command line options.
...
+This behaviour is to remain backwards compatible with older versions of the Lieutenant API.
```
simu, line 43: en_US?
```suggestion
The fallback behavior is in place to remain backwards compatible with older versions of the Lieutenant API.
```

commodore/compile.py
```diff
@@ -25,57 +26,57 @@
...
+    repo_url = cluster.config_repo_url()
     if cfg.debug:
-        click.echo(f" > Cloning customer config {repopath}")
-    repo = git.clone_repository(repopath, P("inventory/classes") / customer_id, cfg)
+        click.echo(f" > Cloning customer config {repo_url}")
+    repo = git.clone_repository(
+        repo_url, P("inventory/classes") / cluster.tenant(), cfg
+    )
+    rev = cluster.config_git_repo_revision()
+    if rev != HEAD:
```
simu, line 58: Same as for `fetch_global_config`.

commodore/compile.py
```diff
@@ -25,57 +26,57 @@
...
 )
 from .postprocess import postprocess_components
 from .refs import update_refs
+from .git import HEAD
 
 
 TARGET = "cluster"
 
 
-def _fetch_global_config(cfg, cluster):
```
simu, line 37: `fetch_global_config` and `fetch_customer_config` are not part of the public API of `commodore.compile`. They can be left as `_fetch_X_config`.

If they need to be part of the public API for some reason, they should be moved out of the module `commodore.compile` to somewhere where they make sense as part of the API (e.g. `commodore.cluster`).

commodore/cluster.py
```diff
@@ -13,14 +13,79 @@
...
-def fetch_cluster(cfg, clusterid):
-    cluster = lieutenant_query(cfg.api_url, cfg.api_token, "clusters", clusterid)
-    # TODO: move Commodore global defaults repo name into Lieutenant
-    # API/cluster facts
-    cluster["base_config"] = "commodore-defaults"
-    return cluster
+from .git import HEAD
+
+
+class Cluster:
```
simu, line 19: Consider adding type hints for the class fields

```suggestion
class Cluster:
    __cfg: Config
    __cluster_response: Dict[str, Any]
    __tenant_response: Dict[str, Any]
```

commodore/cluster.py
```diff
@@ -13,14 +13,79 @@
...
+    def __init__(
+        self,
+        cfg: Config,
+        cluster_response: Dict,
+        tenant_response: Dict,
+    ):
+        self.__cfg = cfg
+        self.__cluster = cluster_response
+        self.__tenant = tenant_response
```
simu, line 28: Instead of saving the raw responses here, consider directly extracting the fields you care about in the constructor.

simu (changes requested):
Change looks good overall, but there's some bits of the implementation that look "odd" to me.

I don't have really strong feelings on any of them though.

commodore/cluster.py
corvus-ch: But then they can also be changed by assigning values right? Maybe I am to conservative here but I like this to not be possible.

commodore/cluster.py
corvus-ch: Conceptually, it would be OK to not have facts. On the other hand, we do have an assertion somewhere else expecting the `cloud` and `distribution` facts. This has partially become obsolete. Once we remove the backwards compatibility layer, Commodore does no longer need any facts. It is up the the hierarchy defined in the global config to define if and which facts are required.

commodore/cluster.py
corvus-ch: I prefer to leave it that way. My first argument is: it keeps functions small. Yes, doing as you suggested would make the getters even smaller but the constructor will be bloated. Second I expect this to be simpler to extend as there is less boiler plate code to change wen new fields are accessed.

commodore/cluster.py
simu: You can't assign to a method with `@property`, you'd need to implement (taking `id` as the example):

```
@id.setter
def set_id(self, id):
    # do whatever is necessary to update the id
    pass
```

See https://docs.python.org/3/library/functions.html#property

corvus-ch: That then raises the question: should all functions of `Cluster` be properties? My gut feelings tell me, this is true for `id`, `tenant` and `facts` but not for the others. But I can not explain why.

simu: From what I can tell, the whole `Cluster` class is a construct which exposes "properties" of the cluster which we're compiling a catalog for, therefore I'd argue that all the current function (except the constructor obviously) could be made properties.

docs/modules/ROOT/pages/reference/cli.adoc
```diff
@@ -33,19 +33,14 @@ This command doesn't have any command line options.
...
+Components are searched in `GLOBAL_GIT_BASE/commodore-components`, unless specified otherwise in the configuration, see xref:reference/architecture.adoc#_component_discovery[Component discovery].
 +
-By default, Commodore also uses the global Git base URL when doing component
-autodiscovery. Components are searched in
-`GLOBAL_GIT_BASE/commodore-components`, unless specified otherwise in the
-configuration, see
-xref:reference/architecture.adoc#_component_discovery[Component discovery].
+If the Lieutenant API doesn't return the global Git repository URL, the global Git base will also be used to determine the URL of that repository.
+The fallback global repository URL is `GLOBAL_GIT_BASE/commodore-defaults.git`.
+The fallback behavior is in place to remain backwards compatible with older versions of the Lieutenant API.
```
simu, line 43:
```suggestion
The fallback behavior is in place to ensure Commodore remains backwards compatible with older versions of the Lieutenant API.
```

simu (approved):
LGTM now (modulo the failing tests).

Thanks for the effort :+1:

PR #227, opened 2020-10-22
Create a Kapitan target for each component

simu:
This PR implements #223.

Notable changes:
* Components are included in the hierarchy via applications instead of classes (64b82ee). Component includes in the commodore-defaults and tenant repos need to be moved to applications, as described in the commit message.
* The new implementation implicitly disallows a component with name "cluster" to exist, as we're using the target name "cluster" for the bootstrapping target, so we can actually discover components, find component versions, generate Kapitan refs, etc. from a well-known target.
* The new implementation requires the name of the Kapitan target to match the name of the component which is being compiled.
* Due to the previous point, the single-component compilation has been adjusted to create a target named the same as the component.

  As a result to this adjustment, single component compilation would overwrite the intermediate result of catalog compilation for the cmponent, if it exists. To separate the result of single component compilation from catalog compilation intermediate results, single-component compilation appends "standalone" to the requested output directory.

## Checklist

- [x] Update tests.


commodore/catalog.py
```diff
@@ -65,15 +66,15 @@ def clean_catalog(repo):
...
     click.secho("Updating catalog repository...", bold=True)
     # pylint: disable=import-outside-toplevel
     from distutils import dir_util
     import textwrap
 
     catalogdir = P(repo.working_tree_dir, "manifests")
-    # copy compiled catalog into catalog directory
-    dir_util.copy_tree(P("compiled") / target_name, str(catalogdir))
+    for target_name in targets:
+        dir_util.copy_tree(str(P("compiled") / target_name), str(catalogdir))
```
srueg, line 77:
```suggestion
        dir_util.copy_tree(str(P("compiled") / target_name), str(catalogdir / target_name))
```
If we change this to copy stuff to a subdir named by the component, we don't have to specify the component name in the various `compile` items anymore:

```
      - output_path: argocd/00_crds/
        input_type: copy
        input_paths:
          - argocd/manifests/${argocd:git_tag}/crds/
```
becomes
```
      - output_path: 00_crds/
        input_type: copy
        input_paths:
          - manifests/${argocd:git_tag}/crds/
```
If we also change the search path to `dependencies/$component_name` the `input_paths` also become nicer.

srueg: It would need some special handling for the `apps` part since all Argo CD apps need to go into the same directory (not by component).

srueg: IMHO it would make components much simpler (and more similar to all the Kapitan examples) but would require some migration path, I guess.
Other opinions?

simu: > If we also change the search path to `dependencies/$component_name` the `input_paths` also become nicer.

This will continue to explode if there's multiple components with identically named files, as we still only invoke `kapitan.compile_targets()` a single time, but now with a list of targets.

srueg: Hmm true. Well in that case we might as well leave it as is. At least it is consistent between `output_path` and `input_paths`.

corvus-ch (approved):
LGTM

Thanks for that work. I am amazed how little has changed. At least from the perspective of Commodores architecture. I am glad I left the target handling. I considered to remove that in #195.

This change brakes existing setups. Do we have a list of follow up tasks. Maybe mention the required changes in the CHANGELOG.

simu:
> This change brakes existing setups. Do we have a list of follow up tasks. Maybe mention the required changes in the CHANGELOG.

Follow-ups are linked in the internal ticket corresponding to #223.

chrisglass (approved):
Looks good! I echo @corvus-ch 's sentiment - it's impressive how little changed.

srueg (approved):

PR #228, opened 2020-10-22
Explain how to do SSH-agent forwarding on macOS

corvus-ch:
## Checklist

### DIFF
--- a/docs/modules/ROOT/pages/running-commodore.adoc
+++ b/docs/modules/ROOT/pages/running-commodore.adoc
@@ -50,7 +50,7 @@ Instead you need to mount as magic path that's provided by Docker for Mac.
 [NOTE]
 ====
 This only works for the stock ssh-agent coming along with macOS.
-If you use any other agent, you are out of luck.
+If you use any other agent, you might be out of luck.
 Docker for mac doesn't support mounting sockets.
 ====
 
--- a/docs/modules/ROOT/pages/running-commodore.adoc
+++ b/docs/modules/ROOT/pages/running-commodore.adoc
@@ -45,7 +45,46 @@ commodore() {
 == macOS X
 
 On macOS with Docker for Mac mounting the SSH agents socket into a container doesn't work.
-Instead your SSH key needs to be mounted into the container.
+Instead you need to mount as magic path that's provided by Docker for Mac.
+
+[NOTE]
+====
+This only works for the stock ssh-agent coming along with macOS.
+If you use any other agent, you are out of luck.
+Docker for mac doesn't support mounting sockets.
+====
+
+[CAUTION]
+====
+That magic socket path belongs to root.
+One must run a container with `--user=0` in order to access it.
+This doesn't mess up your file permissions thanks to the magic of how volume mount work on Docker for Mac.
+====
+
+[source,bash]
+----
+commodore() {
+  docker run \
+    --interactive=true \
+    --tty \
+    --rm \
+    --user="0" \
+    --env COMMODORE_API_URL=$COMMODORE_API_URL \
+    --env COMMODORE_GLOBAL_GIT_BASE=$COMMODORE_GLOBAL_GIT_BASE \
+    --env COMMODORE_API_TOKEN=$COMMODORE_API_TOKEN \
+    --env SSH_AUTH_SOCK=/tmp/ssh_agent.sock \
+    --volume "/run/host-services/ssh-auth.sock:/tmp/ssh_agent.sock" \
+    --volume "${HOME}/.ssh/config:/app/.ssh/config:ro" \
+    --volume "${HOME}/.ssh/known_hosts:/app/.ssh/known_hosts:ro" \
+    --volume "${HOME}/.gitconfig:/app/.gitconfig:ro" \
+    --volume "${PWD}:/app/data" \
+    --workdir /app/data \
+    projectsyn/commodore:latest \
+    $*
+}
+----
+
+Instead you can also mount your SSH key into the container.
 The container will pickup that key and add it do an SSH agent running inside the container.
 You will be prompted to insert your SSH keys password if it has one.
 
@@ -68,7 +107,6 @@ commodore() {
     $*
 }
 ----
-
 [NOTE]
 ====
 If you have multiple SSH keys, you can mount only the one you need for commodore.


docs/modules/ROOT/pages/running-commodore.adoc
```diff
@@ -45,7 +45,47 @@ commodore() {
...
+====
+This only works for the stock ssh-agent coming along with macOS.
+If you use any other agent, you are out of luck.
+Docker for mac doesn't support mounting sockets.
+====
+
+[CAUTION]
+====
+That magic socket path belongs to root.
+One must a container with `--user=0` in order to access it.
```
ccremer, line 60: there seems to be a word missing in `One must [..] a container`.
Btw, "One must ..." is very formal English. You could also just write "You have to [start?] a container..."

docs/modules/ROOT/pages/running-commodore.adoc
```diff
@@ -45,7 +45,47 @@ commodore() {
...
+    --env COMMODORE_API_URL=$COMMODORE_API_URL \
+    --env COMMODORE_GLOBAL_GIT_BASE=$COMMODORE_GLOBAL_GIT_BASE \
+    --env COMMODORE_API_TOKEN=$COMMODORE_API_TOKEN \
+    --env SSH_AUTH_SOCK=/tmp/ssh_agent.sock \
+    --volume "/run/host-services/ssh-auth.sock:/tmp/ssh_agent.sock" \
+    --volume "${HOME}/.ssh/config:/app/.ssh/config:ro" \
+    --volume "${HOME}/.gitconfig:/app/.gitconfig:ro" \
+    --volume "${PWD}:/app/data" \
+    --workdir /app/data \
+    --user=0 \
```
ccremer, line 82: This is probably conflicting/overriding with line 72 (`--user="$(id -u)"`), probably remove line 72?

Also, it's possible to set the group as well with `--user=0:<your-user>`, would something with group membership also solve file permissions?

ccremer (changes requested):

docs/modules/ROOT/pages/running-commodore.adoc
```diff
@@ -45,7 +45,47 @@ commodore() {
 == macOS X
 
 On macOS with Docker for Mac mounting the SSH agents socket into a container doesn't work.
-Instead your SSH key needs to be mounted into the container.
+Instead you need to mount as magic path that's provided by Docker for Mac.
+
+[NOTE]
+====
+This only works for the stock ssh-agent coming along with macOS.
+If you use any other agent, you are out of luck.
```
srueg, line 53: Should we mention that other agents work as well, as long as they allow to store their socket file in a certain directory?

docs/modules/ROOT/pages/running-commodore.adoc
```diff
@@ -45,7 +45,47 @@ commodore() {
...
+  docker run \
+    --interactive=true \
+    --tty \
+    --rm \
+    --user="$(id -u)" \
+    --env COMMODORE_API_URL=$COMMODORE_API_URL \
+    --env COMMODORE_GLOBAL_GIT_BASE=$COMMODORE_GLOBAL_GIT_BASE \
+    --env COMMODORE_API_TOKEN=$COMMODORE_API_TOKEN \
+    --env SSH_AUTH_SOCK=/tmp/ssh_agent.sock \
+    --volume "/run/host-services/ssh-auth.sock:/tmp/ssh_agent.sock" \
```
simu, line 77: Doesn't OSX have `$SSH_AUTH_SOCK` which holds the path to the current ssh-auth socket?

srueg: Yes, but the `/run/host-services/ssh-auth.sock` is a special path inside the Docker VM where the host's agent socket is mounted. So it's always the same path and the host's socket must be at a certain location.

simu: Ah right, there's one more layer of magic on OSX, I forgot that was the original issue. So, to reference the other comment, it should still work with any agent though, since that layer of mapping (host-level `$SSH_AUTH_SOCK` to `/run/host-services/ssh-auth.sock`) is done by Docker on OSX?

srueg: Unfortunately Docker for mac maps a hard coded path into the VM (instead of the generic `$SSH_AUTH_SOCK`).

simu: well, nevermind then

docs/modules/ROOT/pages/running-commodore.adoc
corvus-ch: Do you have a working example? If not, no as it is not confirmed knowledge.

docs/modules/ROOT/pages/running-commodore.adoc
corvus-ch: Yes the word _run_ was missing.

docs/modules/ROOT/pages/running-commodore.adoc
corvus-ch: I tested that one again. I forgot another Docker for Mac magic: your file permissions are always correct :tm:. That is, their owner will always be your user.

I fixed the duplicate `--user` option.

docs/modules/ROOT/pages/running-commodore.adoc
```diff
@@ -45,7 +45,46 @@ commodore() {
 == macOS X
 
 On macOS with Docker for Mac mounting the SSH agents socket into a container doesn't work.
-Instead your SSH key needs to be mounted into the container.
+Instead you need to mount as magic path that's provided by Docker for Mac.
+
+[NOTE]
+====
+This only works for the stock ssh-agent coming along with macOS.
+If you use any other agent, you are out of luck.
```
tobru, line 53:
```suggestion
If you use any other agent, you might be out of luck.
```

docs/modules/ROOT/pages/running-commodore.adoc
srueg: Jup, I actually got it working with my brew-installed ssh-agent without any changes.

tobru (approved):

ccremer (approved):
There are some CI checks failing, maybe have a look before merging. otherwise lgtm

docs/modules/ROOT/pages/running-commodore.adoc
srueg: Ok wait, scratch that. I think it's still starting the macOS agent and not the brew installed one...

srueg (approved):

PR #230, opened 2020-10-26
Implement configurable working directory

corvus-ch:
Implementation for #214

## Checklist

- [x] Keep pull requests small so they can be easily reviewed.
- [x] Update the documentation.
- [x] Update tests.
- [x] Update the ./CHANGELOG.md.
- [x] Link this PR to related issues.

PR #232, opened 2020-10-27
Pass source as single argument to `relsymlink()`

corvus-ch:
#230 starts to get too big to handle. This one got extracted from there.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/dependency_mgmt.py
+++ b/commodore/dependency_mgmt.py
@@ -22,17 +22,16 @@ def create_component_symlinks(cfg, component: Component):
     we want to access some of the file contents through the inventory.
     """
     relsymlink(
-        component.target_directory / "class",
-        f"{component.name}.yml",
-        "inventory/classes/components",
+        component.target_directory / "class" / f"{component.name}.yml",
+        P("inventory/classes/components"),
     )
+
     component_defaults_class = component.target_directory / "class" / "defaults.yml"
     if component_defaults_class.is_file():
         relsymlink(
-            component.target_directory / "class",
-            "defaults.yml",
+            component_defaults_class,
             P("inventory/classes/defaults"),
-            destname=f"{component.name}.yml",
+            dest_name=f"{component.name}.yml",
         )
     else:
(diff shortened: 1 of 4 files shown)


commodore/helpers.py
```diff
@@ -175,14 +175,15 @@ def rm_tree_contents(basedir):
             os.unlink(f)
 
 
-def relsymlink(srcdir, srcname, destdir, destname=None):
-    if destname is None:
-        destname = srcname
+# pylint: disable=unsubscriptable-object
+def relsymlink(src: P, dest_dir: P, dest_name: Union[P, str] = None):
```
simu, line 179: I don't think having `dest_name` as a Path makes sense. I'd suggest making `dest_name` `Optional[str]` to allow `None` or strings.

```suggestion
def relsymlink(src: P, dest_dir: P, dest_name: Optional[str] = None):
```

If the type checker complains about assigning `src.name` to `dest_name` with the adjusted type hint, the assignment can be written as

```
dest_name = str(src.name)
```

commodore/dependency_mgmt.py
```diff
@@ -22,30 +22,30 @@ def create_component_symlinks(cfg, component: Component):
...
+
     component_defaults_class = component.target_directory / "class" / "defaults.yml"
     if component_defaults_class.is_file():
         relsymlink(
-            component.target_directory / "class",
-            "defaults.yml",
+            component_defaults_class,
             P("inventory/classes/defaults"),
-            destname=f"{component.name}.yml",
+            f"{component.name}.yml",
```
simu, line 34: I prefer passing optional arguments using the keyword style, as that tends to avoid unwelcome surprises if the order of the optional arguments changes in the future.

simu (changes requested):

commodore/dependency_mgmt.py
corvus-ch: That was the argument I was missing.

simu (approved):

PR #234, opened 2020-10-27
Add support for component instantiation

### LINKED ISSUE #221: Support for instantiating components multiple times
## Context

Currently Commodore components don't support installing multiple instances of the software they manage. There are already a some components for which installing multiple instances is a common use case, such as [nfs-client-provisioner](https://github.com/projectsyn/component-nfs-client-provisioner/).

### Multi-instance aware components

The "simple" solution from the perspective of Commodore/Kapitan/Reclass is to simply leave multi-instance support to the component authors. For this solution, component authors would explicitly need to implement multi-instance support. This approach would most likely involve a parameter structure which exposes the multi-instance nature of the component in `parameters.<component-name>`.

Taking nfs-client-provisioner as an example, the parameters structure could be something like:

```
parameters:
  nfs_client_provisioner:
    namespace: ...
    common:
      host: ...
    instances:
      instanceA:
        path: ...
      instanceB:
        path: ...
      instanceC:
        host: ...
        path: ...
```

In this example structure, we allow users to define configurations which are shared between multiple (but maybe not all) instances in a key `common`. Instances are configured in key `instances`, and keys in `instances` are used as instance identifiers. Instance configuration overrides configurations in `common`.

The main downside of this approach is that it may not be feasible to implement for components which use Helm charts as their base, unless the Helm chart exposes a similar structure. The reason for this is that Kapitan's Helm templating is expressed in the reclass inventory, and reclass is not flexible enough to instantiate multiple copies of a part of the hierarchy based on another key in the inventory.

### Commodore support for component instantiation

Another approach is to not make components multi-instance aware, but instead implement support for instantiating a component multiple times in Commodore. This approach would potentially scale better, as component authors won't have to reinvent the wheel for every component which they want to support multiple instances. Additionally, Commodore can probably juggle things so that Kapitan's helm templating can be used to instantiate a Helm-chart based component multiple times.

This approach would need logic in Commodore which identifies the component(s) to instantiate multiple times based on the included component classes according to some "instance naming scheme".

Commodore would then have to duplicate the relevant files (component class and defaults mainly), rewriting references to the component name in those files (input/output paths, parameters key, ...) to match the "instance name" which would be derived from the class "instance naming scheme".

Users wishing to instantiate a component multiple times would need to include the component using the "instance naming scheme", and provide parameters for each instance separately. Reclass references can potentially be used to share parameter values between instances.

## Alternative approaches

Another approach which is not outlined in detail above is that when multi-instance support is required, the Commodore component must install an operator which supports instantiating the required software multiple times. This is naturally dependent on an operator existing for the software that needs to be instantiated multiple times, and is similar to the first proposed solution from the perspective of how multiple instances could then be configured via the hierarchy.

## Out of scope

How to manage services for the customer on a Syn-enabled cluster is out of scope for this issue.

simu:
Resolves #221 

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/cluster.py
+++ b/commodore/cluster.py
@@ -123,12 +123,12 @@ def render_target(
     inv: Inventory,
     target: str,
     components: Iterable[str],
-    bootstrap: bool = False,
     # pylint: disable=unsubscriptable-object
     component: Optional[str] = None,
 ):
     if not component:
         component = target
+    bootstrap = target == BOOTSTRAP_TARGET
     if not bootstrap and component not in components:
         raise click.ClickException(f"Target {target} is not a component")
 
@@ -168,22 +168,13 @@ def render_target(
     }
 
 
-def update_target(
-    cfg: Config,
-    target: str,
-    bootstrap=False,
-    # pylint: disable=unsubscriptable-object
-    component: Optional[str] = None,
-):
+# pylint: disable=unsubscriptable-object
+def update_target(cfg: Config, target: str, component: Optional[str] = None):
     click.secho(f"Updating Kapitan target for {target}...", bold=True)
     file = cfg.inventory.target_file(target)
     os.makedirs(file.parent, exist_ok=True)
     targetdata = render_target(
-        cfg.inventory,
-        target,
-        cfg.get_components().keys(),
-        bootstrap=bootstrap,
-        component=component,
+        cfg.inventory, target, cfg.get_components().keys(), component=component
     )
     yaml_dump(targetdata, file)
 
--- a/commodore/cluster.py
+++ b/commodore/cluster.py
@@ -138,8 +138,7 @@ def render_target(
     }
 
     for c in components:
-        defaults_file = inv.defaults_dir / f"{c}.yml"
-        if defaults_file.is_file():
+        if inv.defaults_file(c).is_file():
             classes.append(f"defaults.{c}")
 
     classes.append("global.commodore")
--- a/commodore/cluster.py
+++ b/commodore/cluster.py
@@ -133,7 +133,9 @@ def render_target(
         raise click.ClickException(f"Target {target} is not a component")
 
     classes = [f"params.{BOOTSTRAP_TARGET}"]
-    parameters: Dict[str, Union[Dict, str]] = {}
+    parameters: Dict[str, Union[Dict, str]] = {
+        "_instance": target,
+    }
 
     for c in components:
         defaults_file = inv.defaults_dir / f"{c}.yml"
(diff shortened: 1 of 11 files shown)


commodore/config.py
```diff
@@ -85,3 +90,17 @@ def get_configs(self):
...
+    def get_component_aliases(self):
+        return self._component_aliases
+
+    def register_component_aliases(self, aliases: Dict[str, str]):
+        self._component_aliases = aliases
+
+    def verify_component_aliases(self, cluster_parameters: Dict):
+        for alias, cn in self._component_aliases.items():
+            ckey = cn.replace("-", "_")
+            if "_instance" not in cluster_parameters[ckey]:
```
srueg, line 103: Is the idea that `_instance` should just be an empty dict in order to allow instantiation of a component?

I'd propose to do something a little more explicit, like `parameters.<component-name>.multi_instance: true` for example.
Then we could also move the `_instance` magic parameter directly under `parameters` (like in #241 proposed).

simu: Changed as suggested in the latest version

commodore/cluster.py
```diff
@@ -120,47 +120,71 @@ def read_cluster_and_tenant() -> Tuple[str, str]:
...
     classes = [f"params.{BOOTSTRAP_TARGET}"]
-    parameters = {}
+    parameters: Dict[str, Union[Dict, str]] = {
+        "_instance": target,
+    }
 
-    for component in components:
-        defaults_file = inv.defaults_dir / f"{component}.yml"
+    for c in components:
+        defaults_file = inv.defaults_dir / f"{c}.yml"
```
corvus-ch, line 141: Oh, I must have missed one.

```suggestion
        defaults_file = inv.defaults_file(c)
```

commodore/cluster.py
```diff
@@ -120,47 +120,71 @@ def read_cluster_and_tenant() -> Tuple[str, str]:
 
 
 def render_target(
-    inv: Inventory, target: str, components: Iterable[str], bootstrap=False
+    inv: Inventory,
+    target: str,
+    components: Iterable[str],
+    bootstrap: bool = False,
```
corvus-ch, line 126: Can we get rid of this with `target == BOOTSTRAP_TARGET`?

srueg (approved):
Nice one! 🎉

corvus-ch (approved):

commodore/cluster.py
simu: maybe I messed that one up when rebasing...

PR #237, opened 2020-10-28
Drop check for "Old-style" components

corvus-ch:
## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/dependency_mgmt.py
+++ b/commodore/dependency_mgmt.py
@@ -25,20 +25,11 @@ def create_component_symlinks(cfg, component: Component):
         component.target_directory / "class" / f"{component.name}.yml",
         P("inventory/classes/components"),
     )
-
-    component_defaults_class = component.target_directory / "class" / "defaults.yml"
-    if component_defaults_class.is_file():
-        relsymlink(
-            component_defaults_class,
-            P("inventory/classes/defaults"),
-            dest_name=f"{component.name}.yml",
-        )
-    else:
-        click.secho(
-            "     > Old-style component detected. Please move "
-            + "component defaults to 'class/defaults.yml'",
-            fg="yellow",
-        )
+    relsymlink(
+        component.target_directory / "class" / "defaults.yml",
+        P("inventory/classes/defaults"),
+        dest_name=f"{component.name}.yml",
+    )
 
     libdir = component.target_directory / "lib"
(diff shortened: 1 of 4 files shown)


commodore/dependency_mgmt.py
```diff
@@ -25,20 +25,11 @@ def create_component_symlinks(cfg, component: Component):
...
-            P("inventory/classes/defaults"),
-            dest_name=f"{component.name}.yml",
-        )
-    else:
-        click.secho(
-            "     > Old-style component detected. Please move "
-            + "component defaults to 'class/defaults.yml'",
-            fg="yellow",
-        )
+    relsymlink(
```
srueg, line 28: We should probably still have some check if `class/defaults.yml` exists, otherwise we'll create broken symlinks.

srueg (changes requested):

commodore/dependency_mgmt.py
corvus-ch: I have added a check in `relsymlink`. This probably was a general issue.

simu (approved):

commodore/helpers.py
```diff
@@ -184,6 +184,10 @@ def relsymlink(src: P, dest_dir: P, dest_name: Optional[str] = None):
     # https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to
     link_src = os.path.relpath(src, start=dest_dir)
     link_dst = dest_dir / dest_name
+    if not P(link_src).exists():
```
srueg, line 187: Isn't `link_src` already a Path?
```suggestion
    if not link_src.exists():
```

commodore/helpers.py
```diff
@@ -184,6 +184,10 @@ def relsymlink(src: P, dest_dir: P, dest_name: Optional[str] = None):
...
+        click.ClickException(
+            f"Can not lin {link_src} to {link_dst}. Source does not exist."
```
srueg, line 189:
```suggestion
            f"Can't link {link_src} to {link_dst}. Source does not exist."
```

srueg (changes requested):

commodore/helpers.py
corvus-ch: `relpath` returns type `Text`.

srueg (approved):

PR #239, opened 2020-10-28
More specific properties to Component::target_directory

corvus-ch:
#230 starts to get too big to handle. This one got extracted from there.

## Checklist

### DIFF
--- a/commodore/component/__init__.py
+++ b/commodore/component/__init__.py
@@ -1,4 +1,7 @@
+import os
+
 from pathlib import Path as P
+from typing import Iterable
 
 from git import Repo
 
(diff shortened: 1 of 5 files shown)


commodore/component/__init__.py
```diff
@@ -60,6 +63,27 @@ def version(self, version: str):
     def target_directory(self) -> P:
         return component_dir(self.name)
 
+    @property
+    def class_file(self) -> P:
+        return self.target_directory / "class" / f"{self.name}.yml"
```
simu, line 68: It may be worth considering to always construct these file paths (also applies to `defaults_file`, `lib_files` and `filters_file`) based on `self.repo.working_tree_dir` instead of `self.target_directory`, as `self.target_directory` will change if Commodore changes working directory for any reason.

simu: note that this change would resolve the currently failing `component compile` test for a component with postprocessing filters.

corvus-ch: This makes a whole bunch of testcases to fail. I will revisit this again with #230.

simu: yeah, there's probably a lot of confusion regarding working directory handling in the current code base.

simu (approved):

PR #243, opened 2020-10-28
Fix component compile to not break existing remote in component repo

simu:
Remove unnecessary and wrong reinitialization of the `Component()` object in `commodore component compile`.

Also update the `run component compile` test to verify that the repo remotes are not changed.

Originally noticed in https://github.com/projectsyn/component-espejo/pull/8

## Checklist

- [x] Update tests.


ccremer (approved):
though I can't really judge whether the change makes sense

srueg (approved):

PR #246, opened 2020-10-29
Pull essential libraries with Jsonnet Bundler

### LINKED ISSUE #242: Consider vendoring kube.libsonnet in Commodore
## Context

Standalone component compilation needs to vendor/pull in Bitnami's `kube.libsonnet` somehow if compilation is not executed in a pre-initialized Commodore working directory.

## Proposal

One of
* Vendor `kube.libsonnet` in `commodore/lib`, making some vendored version available through the Commodore package/Docker image.
* Add a static dependency to `kube.libsonnet` in the `Jsonnetfile` which is rendered by Commodore.

## Alternatives

Each component individually vendors `kube.libsonnet` as required.

corvus-ch:
Resolves #242

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/compile.py
+++ b/commodore/compile.py
@@ -1,3 +1,5 @@
+from pathlib import Path as P
+
 import click
 from kapitan.cached import reset_cache as reset_reclass_cache
 from kapitan.resources import inventory_reclass
@@ -20,6 +22,7 @@ from .dependency_mgmt import (
     register_components,
     set_component_overrides,
     write_jsonnetfile,
+    jsonnet_dependencies,
 )
 from .helpers import (
     ApiError,
@@ -136,7 +139,7 @@ def compile(config, cluster_id):
         fetch_jsonnet_libs(config, jsonnet_libs)
 
     if not config.local:
-        write_jsonnetfile(config)
+        write_jsonnetfile(P("jsonnetfile.json"), jsonnet_dependencies(config))
         fetch_jsonnet_libraries()
 
     clean_catalog(catalog_repo)
--- a/commodore/dependency_mgmt.py
+++ b/commodore/dependency_mgmt.py
@@ -1,7 +1,7 @@
 import json
 from pathlib import Path as P
 from subprocess import call  # nosec
-from typing import Dict
+from typing import Dict, Iterable
 
 import click
 
@@ -187,11 +187,10 @@ def fetch_jsonnet_libs(config: Config, libs):
             )
 
 
-def write_jsonnetfile(config: Config):
+def jsonnet_dependencies(config: Config) -> Iterable:
     """
-    Writes the file `jsonnetfile.json` containing all components as local dependencies.
+    Creates a list of Jsonnet dependencies for the given Components.
     """
-
     dependencies = []
 
     for component in config.get_components().values():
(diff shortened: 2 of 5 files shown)


corvus-ch:
Needs some rebasing once those other PRs are done. @simu can you review https://github.com/projectsyn/commodore/pull/246/commits/b93bb62b7638d6f293bb8b14442e23dfe69bec55 for now?

commodore/dependency_mgmt.py
```diff
@@ -243,6 +254,58 @@ def fetch_jsonnet_libraries(cwd: P = None):
             "the jsonnet-bundler executable `jb` could not be found"
         ) from e
 
+    # Link essential libraries for backwards compatibility.
```
srueg, line 257: How is this used? As long as the search path is correct the library should be discovered, no?

commodore/dependency_mgmt.py
```diff
@@ -243,6 +254,58 @@ def fetch_jsonnet_libraries(cwd: P = None):
...
+        cwd / "vendor" / "kube-libsonnet" / "kube.libsonnet", lib_dir, "kube.libjsonnet"
+    )
+    relsymlink(
+        cwd / "vendor" / "component-argocd" / "lib" / "argocd.libjsonnet", lib_dir
+    )
+
+
+def inject_essential_libraries(file: P):
+    """
+    Ensures essential libraries are contained within `jsonnetfile.json`.
```
srueg, line 270:
```suggestion
    Ensures essential libraries are added to `jsonnetfile.json`.
```

commodore/dependency_mgmt.py
```diff
@@ -243,6 +254,58 @@ def fetch_jsonnet_libraries(cwd: P = None):
...
+        has_kube = has_kube or "kube-libsonnet" in remote
+        has_argo = has_argo or "component-argocd" in remote
+
+    if not has_kube:
+        deps.append(
+            {
+                "source": {
+                    "git": {"remote": "https://github.com/bitnami-labs/kube-libsonnet"}
+                },
+                "version": "master",
```
srueg, line 290:
```suggestion
                "version": "v1.14.6",
```

commodore/dependency_mgmt.py
```diff
@@ -243,6 +254,58 @@ def fetch_jsonnet_libraries(cwd: P = None):
...
+    with open(file, "r") as f:
+        data = json.load(f)
+
+    deps = data["dependencies"]
+    has_kube = False
+    has_argo = False
+    for dep in deps:
+        remote = dep.get("source", {}).get("git", {}).get("remote", "")
+        has_kube = has_kube or "kube-libsonnet" in remote
+        has_argo = has_argo or "component-argocd" in remote
```
srueg, line 282: Not sure it makes sense to add the Argo CD library like this. Shouldn't this happen via including the component only?

srueg (changes requested):

commodore/dependency_mgmt.py
```diff
@@ -216,24 +215,36 @@ def write_jsonnetfile(config: Config):
...
+        f.write(json.dumps(data, indent=4))
 
 
-def fetch_jsonnet_libraries(cwd: P = None):
+def fetch_jsonnet_libraries(cwd: P = P(".")):
     """
     Download Jsonnet libraries using Jsonnet-Bundler.
     """
+    jsonnetfile = cwd / "jsonnetfile.json"
+    if not jsonnetfile.exists():
```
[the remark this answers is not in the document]
corvus-ch: No. For the scope of _catalog compile_ the file is explicitly written if not running in _local mode_. For the scope of _component compile_, we do want the use the file provided by the component but create one if there is none.

commodore/dependency_mgmt.py
corvus-ch: No. We include those libraries as `lib/<name>.libjsonnet`. Without those links, we would need to rewrite all components to use `<library name>/<file name>`.

commodore/dependency_mgmt.py
corvus-ch: Define, including the component only?

commodore/compile.py
```diff
@@ -136,7 +139,7 @@ def compile(config, cluster_id):
         fetch_jsonnet_libs(config, jsonnet_libs)
 
     if not config.local:
-        write_jsonnetfile(config)
+        write_jsonnetfile(P("jsonnetfile.json"), jsonnet_dependencies(config))
```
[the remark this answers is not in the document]
corvus-ch: Can I leave this to #230? There I will have this as a property on the config object.

commodore/dependency_mgmt.py
srueg: As in "my cluster includes the Argo CD component".
I realize that this won't work for standalone component compilation. But it kind of feels wrong to me to just download the `master` of the lib if it's now already included... The current approach with faking it at least "breaks" it on purpose.

corvus-ch: Note, the component Argo CD gets added ONLY IF it is not already within the list of dependencies.

So if your component does define a specific version, it will not be changed to `master`.

simu (approved):
LGTM for now. Will probably need some reworks in the future anyway.

commodore/dependency_mgmt.py
corvus-ch: I dropped the Argo Component but kept kube-libsonnet. For the component compile, the Argo library still will be faked.

srueg (approved):

PR #247, opened 2020-10-30
Drop COMMODORE_GLOBAL_GIT_BASE

### LINKED ISSUE #229: Drop COMMODORE_GLOBAL_GIT_BASE
## Context

The variable COMMODORE_GLOBAL_GIT_BASE is no longer used to determine the location of the global config git repository (See #226). However it is still in use for auto discovering component repositories.

Having this discussed this with @simu, we concluded to just drop auto discovery for now. As a consequence, each and every component must be listed in `commodore.yml` within the global config git repo.

Note that the global git base is for now still used for backwards comparability. Commodore is using it when talking to older versions of Lieutenant not providing the URL of the global config git repo.

## Alternatives

Make Lieutenant to provide the value needed for component auto discovery. Change Commodore to make use of that.

corvus-ch:
Resolves #229

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/dependency_mgmt.py
+++ b/commodore/dependency_mgmt.py
@@ -97,21 +97,13 @@ def _read_component_urls(cfg: Config, component_names) -> Dict[str, str]:
             f"Could not read Commodore configuration: {e}"
         ) from e
 
-    for component_name in component_names:
-        component_urls[
-            component_name
-        ] = f"{cfg.default_component_base}/{component_name}.git"
+    for component in commodore_config.get("components", []):
+        if component["name"] in component_names:
+            component_urls[component["name"]] = component["url"]
 
-    if commodore_config is None:
-        click.secho("Empty Commodore config file", fg="yellow")
-    else:
-        for component_override in commodore_config.get("components", []):
-            if cfg.debug:
-                click.echo(
-                    f"   > Found override for component {component_override['name']}:"
-                )
-                click.echo(f"     Using URL {component_override['url']}")
-            component_urls[component_override["name"]] = component_override["url"]
+    for name in component_names:
+        if name not in component_urls:
+            raise click.ClickException(f"No url for component '{name}' configured")
 
     return component_urls
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Removed
 
 * Check for "Old-style" components ([#237])
+* Configuration of the global git base ([#247])
+  All components must be listed in `commodore.yml` within the global configuration repository.
+  The URL of the global configuration repository must be set at Lieutenant on the Tenant object.
 
 ## [v0.3.0] - 2020-10-01
 
@@ -264,4 +267,5 @@ Initial implementation
 [#234]: https://github.com/projectsyn/commodore/pull/234
 [#237]: https://github.com/projectsyn/commodore/pull/237
 [#246]: https://github.com/projectsyn/commodore/pull/246
+[#247]: https://github.com/projectsyn/commodore/pull/247
 [#249]: https://github.com/projectsyn/commodore/pull/249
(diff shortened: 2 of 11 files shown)


commodore/dependency_mgmt.py
```diff
@@ -96,21 +96,13 @@ def _read_component_urls(cfg: Config, component_names) -> Dict[str, str]:
...
-        click.secho("Empty Commodore config file", fg="yellow")
-    else:
-        for component_override in commodore_config.get("components", []):
-            if cfg.debug:
-                click.echo(
-                    f"   > Found override for component {component_override['name']}:"
-                )
-                click.echo(f"     Using URL {component_override['url']}")
-            component_urls[component_override["name"]] = component_override["url"]
+    for component in commodore_config.get("components", []):
```
srueg, line 99: I would switch the loop to iterate through `component_names` and lookup the URL in `commodore_config`. But it doesn't really matter I guess, it's just my preference...

commodore/dependency_mgmt.py
```diff
@@ -96,21 +96,13 @@ def _read_component_urls(cfg: Config, component_names) -> Dict[str, str]:
...
+        if component["name"] in component_names:
+            component_urls[component["name"]] = component["url"]
+
+    for name in component_names:
+        if name not in component_urls:
+            raise click.ClickException(f"No url for component '{name}' configured")
```
srueg, line 105: It would be nice if we could allow to not specify a component in `commodore.yml` but instead use `component_versions` in the hierarchy to set the URL. This would allow to add new components for only a part of the hierarchy (i.e. tenant/cluster specific components).

I think it would be enough to just move this check to after discovering component version overrides.

srueg (changes requested):

commodore/dependency_mgmt.py
srueg: Ok, it's not quite that easy. Do you want to do this in this PR or create a new one/issue?

srueg: It would require iterating `cluster_inventory["parameters"].get("component_versions", None)` to get all URLs from there as well, before cloning the components.

commodore/dependency_mgmt.py
corvus-ch: Actually, it is done this way. But`commodore_config` provides an array. By doing as you suggest, I would need to iterate it for each component in `component_names`. To mitigate this, the input from `commodore_config` is converted to a dict within this loop.

commodore/dependency_mgmt.py
corvus-ch: I consider this outside of the original change scope. Can you please create a follow up issue for this?

commodore/dependency_mgmt.py
```diff
@@ -96,21 +96,13 @@ def _read_component_urls(cfg: Config, component_names) -> Dict[str, str]:
             f"Could not read Commodore configuration: {e}"
         ) from e
 
-    for component_name in component_names:
-        component_urls[
-            component_name
-        ] = f"{cfg.default_component_base}/{component_name}.git"
-
-    if commodore_config is None:
-        click.secho("Empty Commodore config file", fg="yellow")
```
simu, line 105: This should probably be promoted to an error instead of removing the warning.

simu (changes requested):

commodore/dependency_mgmt.py
corvus-ch: It is not removed. It is now an exception. What should be done differently?

srueg (approved):

commodore/dependency_mgmt.py
simu: I don't see an exception for "the file exists, but is empty". Although I guess the final check which throws an exception if there's no URL for a component kinda covers the case where the file exists but is empty.

corvus-ch: Yesterday I was pretty done when writing that comment. Now I get your concern.

Yes, the case of an empty file is no longer covered. I removed it as it does not provide much value (if any). The exception for missing URLs is enough.

simu (approved):

PR #250, opened 2020-10-30
Component checkout improvements

simu:
This PR
* Adds infrastructure to run benchmarks and tests
* Adds tests for `Component.checkout`
* Reimplements `Component.checkout` to be roughly twice as fast as the old implementation.

## Checklist

- [x] Update tests.

### DIFF
--- a/tests/test_component.py
+++ b/tests/test_component.py
@@ -2,10 +2,10 @@ import json
 import pytest
 
 from pathlib import Path as P
-from git import Repo, GitCommandError
+from git import Repo
 
 
-from commodore.component import Component
+from commodore.component import Component, component_dir, RefError
 from commodore.inventory import Inventory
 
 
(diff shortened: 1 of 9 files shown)


tests/test_component.py
```diff
@@ -26,7 +27,124 @@ def _init_repo(tmp_path: P, cn: str, url: str):
...
+        raise ValueError(f"Unknown mode {mode} for test")
+
+    assert local_url in c.repo.remote().urls
+    assert orig_url not in c.repo.remote().urls
+    assert not c.repo.head.is_detached
+    assert c.repo.head.ref.name == "local-branch"
+
+
 def test_init_existing_component(tmp_path: P):
+    os.chdir(tmp_path)
```
corvus-ch, line 147: Drop this when #230is merged first. Otherwise #230 needs to rebase and drop this in commit 'Drop directory changes from tests'.

corvus-ch (approved):
Nice one. Many thanks.

I would like to see those benchmark tests to fail should they become worth. Would this be doable?

srueg (approved):

simu:
> Nice one. Many thanks.
> 
> I would like to see those benchmark tests to fail should they become worth. Would this be doable?

Not sure how to do that, I assume we'd have to keep around the results of old test runs somehow for pytest-benchmark to be able to fail the tests if performance degrades.

PR #253, opened 2020-10-31
Keep dependencies dir on clean up

### LINKED ISSUE #218: Clone components only once
## Context

For each compilation of a cluster, commodore removes all components and clones them again. This is a waste of bandwidth and time.

Instead we could clone them into a separate directory. On compilation, commodore would only clone components not only present. For all the others, only a `git fetch` and `git checkout` will be required.

Making the components available within the inventory can then be done with symlinks.

corvus-ch:
Resolves #218

This one needs #250 to work properly.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/helpers.py
+++ b/commodore/helpers.py
@@ -109,7 +109,8 @@ def clean_working_tree(config: Config):
         rmtree = shutil.rmtree
     click.secho("Cleaning working tree", bold=True)
     rmtree(config.inventory.inventory_dir, ignore_errors=True)
-    rmtree(config.inventory.dependencies_dir, ignore_errors=True)
+    rmtree(config.inventory.lib_dir, ignore_errors=True)
+    rmtree(config.inventory.libs_dir, ignore_errors=True)
     rmtree(config.inventory.output_dir, ignore_errors=True)
     rmtree(config.catalog_dir, ignore_errors=True)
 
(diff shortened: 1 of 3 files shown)


commodore/helpers.py
```diff
@@ -109,7 +109,7 @@ def clean_working_tree(config: Config):
         rmtree = shutil.rmtree
     click.secho("Cleaning working tree", bold=True)
     rmtree(config.inventory.inventory_dir, ignore_errors=True)
-    rmtree(config.inventory.dependencies_dir, ignore_errors=True)
```
srueg, line 112: This has the side-effect that Kapitan won't download HTTP(S) dependencies again if they already exist on disk. Which is ok-ish, I guess.

Should we maybe add a flag to the `cleanup` command to still be able to remove everything?

corvus-ch: No, in this case I just would also drop the `libs` dir. Downloading libs is not what takes time.

srueg: That's not what I meant. Kapitan dependencies are usually downloaded to `dependencies/component-name/manifests/name.yaml` or similar.

simu (approved):

srueg (approved):

srueg:
Unfortunately this breaks if unstaged changes are present in dependencies:
```
cmdline: git checkout master
  stderr: 'error: Your local changes to the following files would be overwritten by checkout:
	class/resource-locker.yml
Please commit your changes or stash them before you switch branches.
Aborting'
```

simu:
Hmm, I didn't cover that case in my `Component.checkout` tests. I assume this error happened when running a `catalog compile` without the `--local` flag.

srueg:
Yes exactly. I think it's actually a nice thing because it prevents local changes from being lost. but maybe the error message could be a little nicer and we could introduce a `--force/-f` flag for example to force a checkout?

simu:
Yes, I agree, we need to properly handle this case and produce a user-friendly error. And a force flag to throw away changes probably makes sense too

PR #269, opened 2020-12-04
Always create targets from component aliases

### LINKED ISSUE #268: Alias-only Instances Not Working as Intended

1. Use a component only with aliases:
  ```yaml
  applications:
    - service-broker as redis
    - service-broker as mariadb
  ```
  2. Compile catalog

## Actual Behavior
The component `service-broker` will be instantiated 3 times with the following names: `redis`, `mariadb` and `service-broker`.

## Expected Behavior
The component should only be instantiated 2 times, with the names `redis` and `mariadb`.

The problematic code is here: https://github.com/projectsyn/commodore/blob/master/commodore/dependency_mgmt.py#L75
and here: https://github.com/projectsyn/commodore/blob/master/commodore/compile.py#L73

Because we add the component to the `components` set and create a target for each component in there, the component will always be instantiated unaliased in addition to all aliased occurrences.

simu:
This commit addresses the issue reported in #268, where we found that if a component is only included in aliased forms, the plain component is rendered to the catalog in addition to the aliased versions.

This commit changes component discovery to handle unaliased components as if they were aliased to the component name itself.

This change in discovery allows us to only generate targets for aliases which correctly doesn't create a target for the unaliased version of a component if it's only included in aliased form.

Fixes #268.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 * Add GitHub action to build Kapitan docker image from `tools/Dockerfile.kapitan` ([#266])
 
+### Changed
+* Always create targets from component aliases ([#269])
+
 ### Fixed
 * Correctly create a local branch when overriding component version ([#262])
 * Field read when determine revision of the tenant config repository ([#263])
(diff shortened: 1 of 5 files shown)


srueg (approved):

srueg:
I tested it on the service-broker component and it works as intended 🎉 
Thanks!

srueg (approved):

PR #275, opened 2021-01-11
 Support for rendering component `jsonnetfile.json` from Jsonnet

### LINKED ISSUE #213: Specify Jsonnet-Bundler Dependency Versions in Hierarchy
## Context
Currently a component can have a `jsonnetfile.json` to specify Jsonnet dependencies. This file is not handled by Kapitan and therefore can't make use of the configuration hierarchy. The only way to have different versions of dependencies is with a separate branch of the component (which then can be specified in the hierarchy).

Implement a possibility to specify Jsonnet dependencies and their version in the hierarchy so they can be overridden.

For example the [kube-prometheus](https://github.com/prometheus-operator/kube-prometheus/) library is specific to the K8s version and therefore needs a different version (git branch/commit) for each K8s version.

## Alternatives
As described, an alternative would be to use different branches to handle this.

simu:
Enable hierarchy-based customization for the `jsonnetfile.json` of components.

As an example, this allows users to specify versions of their jsonnet dependencies in the hierarchy (e.g. kube-prometheus version based on cluster K8s version). See https://github.com/projectsyn/component-rancher-monitoring/pull/1 for an example.

Fixes #213 

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/compile.py
+++ b/commodore/compile.py
@@ -136,8 +136,8 @@ def compile(config, cluster_id):
     # Verify that all aliased components support instantiation after resolving component version overrides.
     config.verify_component_aliases(cluster_parameters)
 
-    for cn, component in config.get_components().items():
-        ckey = cn.replace("-", "_")
+    for component in config.get_components().values():
+        ckey = component.parameters_key
         component.render_jsonnetfile_json(cluster_parameters[ckey])
 
     jsonnet_libs = cluster_parameters.get("commodore", {}).get("jsonnet_libs", None)
--- a/commodore/compile.py
+++ b/commodore/compile.py
@@ -136,6 +136,10 @@ def compile(config, cluster_id):
     # Verify that all aliased components support instantiation after resolving component version overrides.
     config.verify_component_aliases(cluster_parameters)
 
+    for cn, component in config.get_components().items():
+        ckey = cn.replace("-", "_")
+        component.render_jsonnetfile_json(cluster_parameters[ckey])
+
     jsonnet_libs = cluster_parameters.get("commodore", {}).get("jsonnet_libs", None)
     if jsonnet_libs and not config.local:
         fetch_jsonnet_libs(config, jsonnet_libs)
--- a/commodore/component/__init__.py
+++ b/commodore/component/__init__.py
@@ -1,6 +1,7 @@
 from pathlib import Path as P
 from typing import Iterable
 
+import _jsonnet
 import click
 
 from git import Repo, BadName, GitCommandError
@@ -140,6 +141,20 @@ class Component:
             raise RefError(f"Failed to checkout revision '{self.version}'") from e
         except BadName as e:
             raise RefError(f"Revision '{self.version}' not found in repository") from e
+
+    def render_jsonnetfile_json(self, component_params):
+        """
+        Render jsonnetfile.json from jsonnetfile.jsonnet
+        """
+        jsonnetfile_jsonnet = self._dir / "jsonnetfile.jsonnet"
+        if jsonnetfile_jsonnet.is_file():
+            # pylint: disable=c-extension-no-member
+            output = _jsonnet.evaluate_file(
+                str(jsonnetfile_jsonnet),
+                ext_vars=component_params.get("jsonnetfile_parameters", {}),
+            )
+            with open(self._dir / "jsonnetfile.json", "w") as fp:
+                fp.write(output)
 
 
 def component_dir(work_dir: P, name: str) -> P:
--- a/commodore/component/__init__.py
+++ b/commodore/component/__init__.py
@@ -155,7 +155,7 @@ class Component:
         if jsonnetfile_jsonnet.is_file():
             if jsonnetfile_json.name in self.repo.tree():
                 click.secho(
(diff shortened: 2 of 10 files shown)


simu:
Still TODO:
* Render `jsonnetfile.jsonnet` in `commodore component compile`

commodore/compile.py
```diff
@@ -136,6 +136,10 @@ def compile(config, cluster_id):
     # Verify that all aliased components support instantiation after resolving component version overrides.
     config.verify_component_aliases(cluster_parameters)
 
+    for cn, component in config.get_components().items():
+        ckey = cn.replace("-", "_")
```
corvus-ch, line 140: If I am not mistaken, this kind of replace is also done elsewhere. As this is a core concept of Commodore, it might be worth to give this its own dedicated function. This might help in future if we need to change how component names are handled.

commodore/component/__init__.py
```diff
@@ -141,6 +142,20 @@ def checkout(self):
...
+        Render jsonnetfile.json from jsonnetfile.jsonnet
+        """
+        jsonnetfile_jsonnet = self._dir / "jsonnetfile.jsonnet"
+        if jsonnetfile_jsonnet.is_file():
+            # pylint: disable=c-extension-no-member
+            output = _jsonnet.evaluate_file(
+                str(jsonnetfile_jsonnet),
+                ext_vars=component_params.get("jsonnetfile_parameters", {}),
+            )
+            with open(self._dir / "jsonnetfile.json", "w") as fp:
```
corvus-ch, line 156: What if both `jsonnetfile.jsonnet` and `jsonnetfile.json` exist (and are checked in into git)? Can we print a warning for that case?

commodore/component/compile.py
```diff
@@ -96,6 +96,12 @@ def compile_component(
                 )
             )
 
+        # Render jsonnetfile.jsonnet if necessary
+        nodes = inventory_reclass(inv.inventory_dir)["nodes"]
+        component_params = nodes[component_name]["parameters"].get(
+            component_name.replace("-", "_"), {}
```
corvus-ch, line 102: Ah yes, and here it is again.

docs/modules/ROOT/pages/reference/concepts.adoc
```diff
@@ -45,6 +45,12 @@ inventory to install the software.
 Components can optionally provide Jsonnet libraries which can be used by other components.
 To allow Commodore to make component libraries available to other components, they must be placed in directory `lib/`.
 
+Components can specify Jsonnet dependencies in a `jsonnetfile.json` in the component's root directory.
+Commodore uses https://github.com/jsonnet-bundler/jsonnet-bundler[jsonnet-bundler] to fetch all components' dependencies.
+
+Optionally, components can provide a `jsonnetfile.jsonnet` instead which will be rendered to JSON before Commodore fetches component dependencies.
+In this case, external variables which are used in the `jsonnetfile.jsonnet` must be provided in the inventory key `parameters.<component_name>.jsonnetfile_parameters`.
```
corvus-ch, line 52: What is the reason behind restricting to a hard coded parameter key?

docs/modules/ROOT/pages/reference/concepts.adoc
```diff
@@ -45,6 +45,12 @@ inventory to install the software.
...
+
```
corvus-ch, line 53: Please add an example of a `jsonnetfile.jsonnet` file.

corvus-ch (changes requested):
As a user it usually makes things easier when having a working example. Please add one to the documentation.

Implementation wise, this is looking good to me. However I wonder why restricting to a hard coded parameter key.

docs/modules/ROOT/pages/reference/concepts.adoc
srueg: Good point, I also don't see a reason to restrict this?

simu: Jsonnet expects a flat dict of strings to strings for the external variables. Therefore we cannot directly pass in the component's parameters dict, as that dict can contain any values.

docs/modules/ROOT/pages/reference/architecture.adoc
```diff
@@ -156,6 +156,57 @@ Similar to Helm charts, the components themselves must make sure to not cause an
...
+
+Commodore injects the key `parameters.<component_name>.jsonnetfile_parameters` as external variables when rendering the `jsonnetfile.jsonnet`.
+
+[NOTE]
+====
+Jsonnet external variables must be string-valued. Therefore we can't simply pass the full `parameters.component_name` as external variables.
+====
+
+Below a `jsonnetfile.jsonnet` and corresponding `class/defaults.yml` for component `rancher-monitoring` are shown.
+The `rancher-monitoring` component on the `kube-prometheus` Jsonnet library, but requires different versions of the library depending on the target cluster's Kubernetes version.
```
corvus-ch, line 175:
```suggestion
The `rancher-monitoring` component depends on the `kube-prometheus` Jsonnet library, but requires different versions of the library depending on the target cluster's Kubernetes version.
```

corvus-ch (approved):
Well, that is a whole other level. Thanks.

srueg (approved):

PR #281, opened 2021-02-08
Generate a best-effort SSH-based push URL for component repos

simu:
This change assumes that most Git hosting services will use the pattern `ssh://git@example.com/path/to/repo.git` for SSH-based push URLs.

We configure each component repo's push-URL using the assumed substitution. This may not always generate a working push-URL, but for popular services like GitHub and GitLab, this removes the need of manually fixing the origin remote when trying to push a feature branch.

## Checklist

- [x] Update tests.

### DIFF
--- a/tests/test_component.py
+++ b/tests/test_component.py
@@ -69,6 +69,7 @@ def test_component_setup_remote_no_sub(tmp_path, repo_url):
         ("http://host/path/to/repo.git", "ssh://git@host/path/to/repo.git"),
         ("https://host/path/to/repo.git", "ssh://git@host/path/to/repo.git"),
         ("https://user@host/path/to/repo.git", "ssh://git@host/path/to/repo.git"),
+        ("https://user:pass@host/path/to/repo.git", "ssh://git@host/path/to/repo.git"),
     ],
 )
 def test_component_setup_remote_sub(tmp_path, repo_url, push_url):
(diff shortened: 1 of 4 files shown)


srueg (approved):

tests/test_component.py
```diff
@@ -43,12 +46,52 @@ def _setup_component(
...
+    assert push_remote == repo_url
+
+
+@pytest.mark.parametrize(
+    "repo_url,push_url",
+    [
+        ("http://host/path/to/repo.git", "ssh://git@host/path/to/repo.git"),
+        ("https://host/path/to/repo.git", "ssh://git@host/path/to/repo.git"),
+        ("https://user@host/path/to/repo.git", "ssh://git@host/path/to/repo.git"),
+        ("https://user:pass@host/path/to/repo.git", "ssh://git@host/path/to/repo.git"),
```
corvus-ch, line 72: What happens to `http://example.com:1234/path/to/repo.git` or http://user:pass@example.com:1234/path/to/repo.git?

corvus-ch (changes requested):
There is an edge case not covered by the tests. Please update.

tests/test_component.py
simu: I've refactored the implementation to drop everything but the host and path of the input URL, if the url starts with `http://` or `https://`.

Obviously we can't guess a potential non-standard SSH port for components which are served on a non-standard HTTP(S) port.

corvus-ch (approved):

PR #284, opened 2021-03-03
 Move component URLs to parameters.components

### LINKED ISSUE #254: Unify how component URLs are defined
## Context

#247 dropped auto discover of components. This makes it mandatory to have all required components listed in `commodore.yml` (name and URL). It is also possible to define component URLs within the hierarchy. This duplication should be removed. We should have only one place to have those URLs defined/configured.

simu:
This commit removes the `components` list in `commodore.yml` in the global config repository and instead looks for component URLs in `parameters.components`.

We deprecate `parameters.component_versions`, but to allow gradual migration to `parameters.components`, we merge
`parameters.component_versions` into `parameters.components` in the Kapitan targets.

Resolves #254 

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/dependency_mgmt.py
+++ b/commodore/dependency_mgmt.py
@@ -2,17 +2,14 @@ import os
 import json
 from pathlib import Path as P
 from subprocess import call  # nosec
-from typing import Dict, Iterable, Optional, Tuple
+from typing import Dict, Iterable, List, Optional, Tuple
 
 import click
-
-from kapitan.cached import reset_cache as reset_reclass_cache
-from kapitan.resources import inventory_reclass
 
 from . import git
 from .config import Config
 from .component import Component, component_dir
-from .helpers import relsymlink, delsymlink
+from .helpers import relsymlink, delsymlink, kapitan_inventory
 
 
 def create_component_symlinks(cfg, component: Component):
@@ -51,16 +48,14 @@ def delete_component_symlinks(cfg, component: Component):
         delsymlink(file, cfg.debug)
 
 
-def _discover_components(cfg, inventory_path):
+def _discover_components(cfg) -> Tuple[List[str], Dict[str, str]]:
     """
     Discover components in `inventory_path/` by extracting all entries from
     the reclass applications dictionary.
     """
-    reset_reclass_cache()
-    inv = inventory_reclass(inventory_path)
-    kapitan_applications = inv["applications"]
+    kapitan_applications = kapitan_inventory(cfg, key="applications")
     components = set()
-    component_aliases = {}
+    component_aliases: Dict[str, str] = {}
     for component in kapitan_applications.keys():
         try:
             cn, alias = component.split(" as ")
@@ -90,9 +85,8 @@ def _read_components(
     component_urls = {}
     component_versions = {}
 
-    reset_reclass_cache()
-    inv = inventory_reclass(cfg.inventory.inventory_dir)
-    cluster_inventory = inv["nodes"][cfg.inventory.bootstrap_target]
+    inv = kapitan_inventory(cfg)
+    cluster_inventory = inv[cfg.inventory.bootstrap_target]
     components = cluster_inventory["parameters"].get("components", None)
     if not components:
         raise click.ClickException("Component list ('parameters.components') missing")
@@ -135,9 +129,7 @@ def fetch_components(cfg: Config):
 
     click.secho("Discovering components...", bold=True)
     cfg.inventory.ensure_dirs()
-    component_names, component_aliases = _discover_components(
-        cfg, cfg.inventory.inventory_dir
-    )
+    component_names, component_aliases = _discover_components(cfg)
     click.secho("Registering component aliases...", bold=True)
     cfg.register_component_aliases(component_aliases)
     urls, versions = _read_components(cfg, component_names)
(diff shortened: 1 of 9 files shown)


commodore/dependency_mgmt.py
```diff
@@ -84,27 +84,43 @@ def _discover_components(cfg, inventory_path):
...
 
-    if cfg.debug:
-        click.echo(f"   > Read commodore config file {cfg.config_file}")
-    try:
-        commodore_config = yaml_load(cfg.config_file)
-    except Exception as e:
-        raise click.ClickException(
-            f"Could not read Commodore configuration: {e}"
-        ) from e
+    reset_reclass_cache()
```
srueg, line 93: We should probably provide a helper function to get the inventory and make sure reset cache is called.

commodore/dependency_mgmt.py
```diff
@@ -84,27 +84,43 @@ def _discover_components(cfg, inventory_path):
...
+    components = cluster_inventory["parameters"].get("components", None)
+    if not components:
+        raise click.ClickException("Component list (parameters.components) missing")
 
-    for component in commodore_config.get("components", []):
-        if component["name"] in component_names:
-            component_urls[component["name"]] = component["url"]
+    for component_name in component_names:
+        if component_name not in components:
+            raise click.ClickException(f"Unknown component '{component_name}'")
```
srueg, line 102:
```suggestion
            raise click.ClickException(f"Unknown component '{component_name}'. Please add it to 'parameters.components'.")
```

commodore/dependency_mgmt.py
```diff
@@ -84,27 +84,43 @@ def _discover_components(cfg, inventory_path):
...
+                f"No url for component '{component_name}' configured"
+            )
+
+        component_urls[component_name] = info["url"]
+        if cfg.debug:
+            click.echo(f" > URL for {component_name}: {component_urls[component_name]}")
+        if "version" in info:
+            component_versions[component_name] = info["version"]
+        else:
+            component_versions[component_name] = "master"
```
srueg, line 117: Would it be possible to use `origin/HEAD` instead of hard coding `master` as the branch name?

srueg: We might even think about making the version required instead of defaulting to `master`. This would require people to think about the version they want to install? Or maybe log a warning and at some point make it required?

docs/modules/ROOT/pages/reference/architecture.adoc
```diff
@@ -36,20 +36,26 @@ applications:
...
-Instead all components which are referenced in the hierarchy must be listed in key `components` in the file `commodore.yml` in the root of the global configuration repository.
-The key `components` holds an array of dictionaries with keys `name` and `url`.
-Each array entry specifies the repository URL of a single component.
-A sample `commodore.yml` can be found below.
-The file `commodore.yml` is also used to define the xref:commodore:ROOT:reference:hierarchy.adoc[Commodore class hierarchy].
+Instead all components which are referenced in the `applications` array must be listed in key `parameters.components` in the hierarchy.
+Using the xref:commodore:ROOT:reference/concepts.adoc#_configuration_hierarchy[configuration hierarchy] for specifying component locations and versions allows tight integration of component management with the rest of the configuration.
+
+Commodore will read `parameters.components` from the hierarchy _before component defaults are included_.
+The key `parameters.components` holds an dictionary of dictionaries mapping component names to their remote repository location and version.
```
srueg, line 43:
```suggestion
The key `parameters.components` holds a dictionary of dictionaries, mapping component names to their remote repository location and version.
```

docs/modules/ROOT/pages/reference/architecture.adoc
```diff
@@ -36,20 +36,26 @@ applications:
...
-  url: https://github.com/projectsyn/component-argocd.git
-- name: metrics-server
-  url: https://github.com/projectsyn/component-metrics-server.git
+parameters:
+  components:
+    argocd:
+      url: https://github.com/projectsyn/component-argocd.git
+    metrics-server:
+      url: https://github.com/projectsyn/component-metrics-server.git
+      version: master
```
srueg, line 58:
```suggestion
      version: v1.0.2
```

srueg (changes requested):

commodore/dependency_mgmt.py
simu: > Would it be possible to use `origin/HEAD` instead of hard coding `master` as the branch name?

I've already got a follow-up PR for using `origin/HEAD` by default instead of master. I decided to split the two to keep the review simpler.

> We might even think about making the version required instead of defaulting to `master`. This would require people to think about the version they want to install? Or maybe log a warning and at some point make it required?

That might be a good idea, especially since we're considering moving to versioned components.

Implementation-wise, I was thinking about building some "deprecation notice" framework yesterday. I was thinking about warning about uses of `parameters.component_versions`, but deprecating components without versions might be a good use case too.

commodore/helpers.py
```diff
@@ -163,6 +164,14 @@ def kapitan_compile(
     )
 
 
+def kapitan_inventory(config: Config) -> Dict:
+    """
+    Reset reclass cache and render inventory
+    """
+    reset_reclass_cache()
+    return inventory_reclass(config.inventory.inventory_dir)
```
srueg, line 172: Shouldn't this return the "nodes" value?
```suggestion
    return inventory_reclass(config.inventory.inventory_dir)["nodes"]
```

simu: Not with the current implementation of `dependency_mgmt._discover_components()` but I've now refactored that to use the `applications` array of target `cluster` instead of the top-level `applications` dictionary keys

srueg (approved):

corvus-ch (approved):

PR #288, opened 2021-03-08
Explicitly set multiprocessing method

srueg:
To avoid different OS defaults.

## Checklist


commodore/cli.py
```diff
@@ -256,5 +257,8 @@ def component_compile(config: Config, path, values, search_paths, output, verbos
 
 
 def main():
+    # set 'fork' method as a more deterministic/conservative
+    # and compatible multiprocessing method for Linux and MacOS
+    multiprocessing.set_start_method("fork")
```
simu, line 262: From https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods:

> Changed in version 3.8: On macOS, the spawn start method is now the default. The fork start method should be considered unsafe as it can lead to crashes of the subprocess. See [bpo-33725](https://bugs.python.org/issue33725).

Not sure if that's something to consider. On a related note, I'm able to replicate the crash on Linux by explicitly calling `multiprocessing.set_start_method("spawn")`

srueg: Good point. But reading through the issue I can't really tell if it's something to worry about. Also since Kapitan decided it's safe I don't see a reason not to do it either.

simu: Good enough for me

simu (approved):

commodore/cli.py
Jean-Daniel: Not sure this was the right decision on the Kapitan part: https://github.com/kapicorp/kapitan/issues/704.

You really should never use fork on macOS.

PR #316, opened 2021-05-07
Implement support for deprecating components

simu:
This PR introduces a new component parameter `_metadata` which is defined as constant in the component template. When adding the parameter to existing components users are responsible to ensure that they mark the parameter as constant (with the `=` prefix).

Further, this PR changes Commodore will print out deprecation notices for all components for which `parameters.<component-name>._metadata.deprecated` is `True`.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/config.py
+++ b/commodore/config.py
@@ -168,6 +168,9 @@ class Config:
     def print_deprecation_notices(self):
         tw = textwrap.TextWrapper(
             width=100,
+            # Next two options ensure we don't break URLs
+            break_long_words=False,
+            break_on_hyphens=False,
             initial_indent=" > ",
             subsequent_indent="   ",
         )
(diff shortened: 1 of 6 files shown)


commodore/config.py
```diff
@@ -176,3 +179,16 @@ def print_deprecation_notices(self):
...
+    def register_component_deprecations(self, cluster_parameters):
+        for cname in self._component_aliases.values():
+            ckey = component_parameters_key(cname)
+            cmeta = cluster_parameters[ckey].get("_metadata", {})
+
+            if cmeta.get("deprecated", False):
+                msg = f"Component {cname} is deprecated."
+                if "replaced_by" in cmeta:
+                    msg += f" Use component {cmeta['replaced_by']} instead."
+                if "deprecation_guide" in cmeta:
```
ccremer, line 192: Maybe rename to parameter to `migration_guide`?

ccremer (approved):
Overall LGTM

commodore/config.py
simu: I had it like that previously, but it seemed more ambiguous, migration guide could be used to point to migration guides for major component upgrades too.

ccremer: True.
It occurred to me that maybe there's not always a replacement available, so it may not always be a guide.
Maybe `deprecation_announcement` or `deprecation_notice_link` something along this way ?

simu: Hm, or just a free-form `deprecation_notice` which gets printed verbatim?

ccremer (approved):

PR #328, opened 2021-06-16
Fix releases to show "x commits to master since this release"

simu:
This notice is only shown when the release is created from a branch, and not if it's created from a commit id or tag, see [this discussion].

This PR ensures that all releases are created from branch "master" and therefore have the message showing how many commits have been added to "master" since the release.

[this discussion]: https://github.community/t/x-commits-to-branch-since-this-release-is-missing-part-2/1806

## Checklist

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -46,11 +46,10 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     - name: Create Release
       if: startsWith(github.ref, 'refs/tags/v')
-      uses: actions/create-release@v1
+      uses: ncipollo/release-action@v1
       with:
-        tag_name: ${{ github.ref }}
-        release_name: ${{ github.ref }}
         body: ${{steps.build_changelog.outputs.changelog}}
         prerelease: "${{ contains(github.ref, '-rc') }}"
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # Ensure target branch for release is "master"
+        commit: master
+        token: ${{ secrets.GITHUB_TOKEN }}


ccremer:
Actually, I've just noticed that https://github.com/actions/create-release is now unmaintained...

Maybe we should find a new GH action instead?

simu:
I saw that too, but I didn't want to invest a lot of time evaluating the alternatives right now

simu:
Switched to the ncipollo action, as there hasn't been a release for the softprops action since early 2020 (and there's open security fix PRs).

.github/workflows/push.yml
```diff
@@ -46,11 +46,11 @@ jobs:
...
       with:
-        tag_name: ${{ github.ref }}
-        release_name: ${{ github.ref }}
+        name: ${{ env.VERSION }}
         body: ${{steps.build_changelog.outputs.changelog}}
         prerelease: "${{ contains(github.ref, '-rc') }}"
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # Ensure target branch for release is "master"
+        commit: master
```
ccremer, line 55: Are you sure this is what you wanted?
The readme says
> commit: An optional commit reference. This will be used to create the tag if it does not exist.

.github/workflows/push.yml
```diff
@@ -46,11 +46,11 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     - name: Create Release
       if: startsWith(github.ref, 'refs/tags/v')
-      uses: actions/create-release@v1
+      uses: ncipollo/release-action@v1
       with:
-        tag_name: ${{ github.ref }}
-        release_name: ${{ github.ref }}
+        name: ${{ env.VERSION }}
```
ccremer, line 51: I don't think the name's necessary
> name: An optional name for the release. If this is omitted the tag will be used.
> tag: An optional tag for the release. If this is omitted the git ref will be used (if it is a tag).

.github/workflows/push.yml
simu: I checked the implementation to ensure the value is used to specify the target for the release as discussed in the linked community post, and testing on my fork shows that it works as intended: https://github.com/simu/commodore/releases/tag/v0.7.1

.github/workflows/push.yml
simu: Checked again, name isn't required even though earlier testing suggested it was.

ccremer (approved):

PR #334, opened 2021-06-24
component-template: Add 'main' as default branch

mhutter:
Most tools and platforms are swaying towards using `main` as the default
branch name. This change makes sure the template works with 'main' as
the default branch out of the box, while retaining backwards
compatibility.

Unfortunately it is not possible to express "default branch" in a GitHub
action.

**Note:** I would also be fine with removing `master` entirely.

## Checklist

- [ ] ~Update the documentation.~
- [ ] ~Update tests.~
- [ ] ~Link this PR to related issues.~

### DIFF
--- a/commodore/component-template/{{ cookiecutter.slug }}/.github/workflows/test.yaml
+++ b/commodore/component-template/{{ cookiecutter.slug }}/.github/workflows/test.yaml
@@ -2,6 +2,7 @@ name: Pull Request
 on:
   pull_request:
     branches:
+      - main
       - master
 
 jobs:


ccremer:
While I have no objection against supporting `master` and `main` branches, most tools cannot handle both.
E.g. the upcoming modulesync can handle `main` branch, but it expects that ALL managed components will have their branch named the same. So either all component use `master` OR `main`, but there can't be a mix.

mhutter:
Tools should not make assumptions about the name of the default branch. It can be determined either via git (`git remote show origin | sed -n '/HEAD branch/s/.*: //p'`) or by querying the GitHub API ([field `default_branch`](https://docs.github.com/en/rest/reference/repos#get-a-repository))

ccremer:
> Tools should not make assumptions about the name of the default branch. It can be determined either via git (`git remote show origin | sed -n '/HEAD branch/s/.*: //p'`) or by querying the GitHub API ([field `default_branch`](https://docs.github.com/en/rest/reference/repos#get-a-repository))

I'm with you. In general, reality shows a different picture though. Some software are years old...
IMO I'm fine with completely renaming all default branches to `main`. It's not that much of an effort, but something to be aware of.
Renaming to `main` also lowers the boundaries to add new components to modulesync (think one "checkbox" reduced)

commodore/component-template/{{ cookiecutter.slug }}/.github/workflows/test.yaml
```diff
@@ -2,6 +2,7 @@ name: Pull Request
 on:
   pull_request:
     branches:
+      - main
       - master
```
ccremer, line 6:
```suggestion
      - main
```

ccremer (approved):

simu:
I don't really care one way or the other re `main` vs `master`.

However, please remember that Commodore creates the component repository and initial commit of a new component as part of `commodore component new`, see  see https://github.com/projectsyn/commodore/blob/65dc21cf03f26ab103efe76fc63c3c4fbf757b3d/commodore/component/template.py#L86-L100

A quick read of the git-python docs (https://gitpython.readthedocs.io/en/stable/reference.html#git.repo.base.Repo.init) and `man git-init` didn't show a way to override the default branch name when calling `git init`. The consensus seems to be that one needs to set the default branch with `git config --global init.defaultBranch main`.

This means that currently Commodore uses the local Git configuration to determine what the default branch for a new component should be, and this may result in components having a different default branch depending on the Git configuration of the user who created them.

If we want a smooth experience when developing components, we probably want to change Commodore to ensure that the component default branch is always the same.

ccremer:
Could commodore rename the master branch to main after it has been initiated?

simu:
> Could commodore rename the master branch to main after it has been initiated?

Sure, if someone implements it

mhutter:
Yes, this is actually how I first stumbled upon the issue since I have `init.defaultBranch` set to `main` on my machine.

Changing the default branch is relatively easy as long as you did not push code to any remote yet (`git branch -M newbranchname`).

On GitHub it's possible to change the default branch via the settings page of the repository (or via the API), I assume the same is true for GitLab and others.

In the long run I think it makes sense to go over the commodore code again to make sure that it does not make assumption about default branch names, but that's out of scope for this PR.

simu:
When cloning components, Commodore already has code to figure out the default branch of the remote repo, cf. https://github.com/projectsyn/commodore/blob/65dc21cf03f26ab103efe76fc63c3c4fbf757b3d/commodore/component/__init__.py#L126-L135

When cloning the global/tenant config repo, Commodore essentially just calls out to `git clone` and doesn't specify `master` anywhere.

simu:
Closing this PR since we've moved the component template out of the Commodore code base into its own repository. New issue to track this request on the template repo: https://github.com/projectsyn/commodore-component-template/issues/13

PR #336, opened 2021-06-25
Use current working directory when searching for .env

laserb:
The `.env` was only found if it is placed next to the virtualenv.
If the virtualenv is in another directory, then the `.env` was ignored and `commodore` failed.
Fix it by searching the `.env` in current directory.

## Checklist

### DIFF
--- a/commodore/cli.py
+++ b/commodore/cli.py
@@ -2,7 +2,7 @@ from pathlib import Path
 
 import click
 
-from dotenv import load_dotenv
+from dotenv import load_dotenv, find_dotenv
 from importlib_metadata import version
 from commodore import __git_version__
 from .catalog import catalog_list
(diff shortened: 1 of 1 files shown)


simu (approved):
LGTM.

Just to clarify the issue description: from what I can tell, the issue doesn't stem from the relative location of the virtualenv in respect to the `.env` file but rather the relative location of the Commodore package compared to the `.env` file, since the old implementation works fine for me with `.env` in `/home/simon/work/syn/commodore` and the Poetry virtualenv in `/home/simon/.cache/pypoetry/virtualenvs/commodore-mfZDsTMQ-py3.9`.

laserb:
Yes, the relative location of the `.env` and the `commodore.cli` is important. `dotenv` searches in the parent folders of `commodore.cli`. It fails if you do not run it with `poetry run`. I.e. install it with `poetry install` or `pip`.

PR #337, opened 2021-06-25
Create a pypi package and publish it

laserb:
Create and publish a pypi package.
It contains pre-compiled helm bindings. They need to be copied to the right place with `copy_helm_bindings` after the pip install.

Package is only build on master and pushed on tags.

## Todo

[ ] Pushing to pypi needs a token in the github repo.

## Checklist

- [x]  Update documentation

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -32,6 +32,8 @@ jobs:
           mkdir -p commodore/helm_bindings/dependency_manager
           cp $(poetry env info -p)/lib/python3.8/site-packages/kapitan/inputs/helm/{helm_binding.py,libtemplate.so} commodore/helm_bindings/inputs
           cp $(poetry env info -p)/lib/python3.8/site-packages/kapitan/dependency_manager/helm/{helm_fetch_binding.py,helm_fetch.so} commodore/helm_bindings/dependency_manager
+      - name: inject version
+        run: make inject-version
       - name: build commodore
         run: poetry build
       - name: publish to pypi
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -1,0 +1,39 @@
+name: PyPI release
+
+on:
+  push:
+    branches:
+    - master
+
+jobs:
+  build-and-publish:
+    name: Build and publish to PyPI
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: install poetry
+        run: pip install poetry
+      - name: install dependencies
+        run: poetry install
+      - name: install go
+        run: |
+          sudo apt update
+          sudo apt install golang-1.14-go -y
+      - name: build helm bindings
+        run: |
+          bash $(poetry env info -p)/lib/python3.8/site-packages/kapitan/inputs/helm/build.sh
+          bash $(poetry env info -p)/lib/python3.8/site-packages/kapitan/dependency_manager/helm/build.sh
+          mkdir -p commodore/helm_bindings/inputs
+          mkdir -p commodore/helm_bindings/dependency_manager
+          cp $(poetry env info -p)/lib/python3.8/site-packages/kapitan/inputs/helm/{helm_binding.py,libtemplate.so} commodore/helm_bindings/inputs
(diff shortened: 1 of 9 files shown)


pyproject.toml
```diff
@@ -1,13 +1,20 @@
 [tool.poetry]
-name = "commodore"
-version = "v0.0.0"
+name = "syn-commodore"
+version = "v0.7.0"
```
simu, line 3: Please don't hardcode the version in `pyproject.toml`, but rather generate the final version from the Git tag/commit in the Github action, see https://github.com/projectsyn/commodore/blob/8671d53ef7e9fb1b613b8d0f538bb9411edb5588/Dockerfile#L25-L30 for an example.

Use https://github.com/projectsyn/commodore/blob/8671d53ef7e9fb1b613b8d0f538bb9411edb5588/Makefile#L44-L45 to generate the required values for `GITVERSION` and `PYVERSION`

.github/workflows/publish-pypi.yml
```diff
@@ -0,0 +1,39 @@
...
+          sudo apt update
+          sudo apt install golang-1.14-go -y
+      - name: build helm bindings
+        run: |
+          bash $(poetry env info -p)/lib/python3.8/site-packages/kapitan/inputs/helm/build.sh
+          bash $(poetry env info -p)/lib/python3.8/site-packages/kapitan/dependency_manager/helm/build.sh
+          mkdir -p commodore/helm_bindings/inputs
+          mkdir -p commodore/helm_bindings/dependency_manager
+          cp $(poetry env info -p)/lib/python3.8/site-packages/kapitan/inputs/helm/{helm_binding.py,libtemplate.so} commodore/helm_bindings/inputs
+          cp $(poetry env info -p)/lib/python3.8/site-packages/kapitan/dependency_manager/helm/{helm_fetch_binding.py,helm_fetch.so} commodore/helm_bindings/dependency_manager
```
simu, line 34: This will not produce a usable package for users on MacOS.

commodore/cli.py
```diff
@@ -2,7 +2,7 @@
 
 import click
 
-from dotenv import load_dotenv
+from dotenv import load_dotenv, find_dotenv
 from importlib_metadata import version
 from commodore import __git_version__
```
simu, line 7:
```suggestion
from commodore import __git_version__, __version__
```

commodore/cli.py
```diff
@@ -23,7 +23,7 @@
 
 
 def _version():
-    pyversion = version("commodore")
+    pyversion = version("syn-commodore")
     if f"v{pyversion}" != __git_version__:
         return f"{pyversion} (Git version: {__git_version__})"
     return pyversion
```
simu, line 29:
```suggestion
    if f"v{__version__}" != __git_version__:
        return f"{__version__} (Git version: {__git_version__})"
    return __version__
```

commodore/copy_helm_bindings.py
```diff
@@ -0,0 +1,36 @@
...
+
+    for f in ["helm_fetch_binding.py", "helm_fetch.so"]:
+        shutil.copyfile(
+            os.path.join(dependency_manager_source_dir, f),
+            os.path.join(dependency_manager_target_dir, f),
+        )
+
+
+if __name__ == "__main__":
+    main()
```
simu, line 36: Probably worth to print a warning on MacOS instead of silently installing non-working ELF libraries.

simu (commented):
Looks good so far, thanks for the implementation.

I left some inline comments regarding the package version configuration.

I'm not feeling too great about building the Kapitan Helm bindings and distributing them as part of Commodore, but hopefully this won't be needed for much longer.

.github/workflows/publish-pypi.yml
laserb: They just cannot use the pre-compiled helm-bindings. Compiling them after the `pip install` still works.

laserb: But compiling will fail with newer version of go, because kapitan did not release this fix yet: https://github.com/kapicorp/kapitan/commit/a996ea7cda9a1ea863c8462c98e8c9cad8a11480

commodore/copy_helm_bindings.py
laserb: We added an error message.

.github/workflows/publish-pypi.yml
simu: The next version of Kapitan won't provide the Helm bindings at all anymore, and will instead call out to the Helm binary.

simu (approved):
LGTM

PR #341, opened 2021-06-29
Always fetch Kapitan dependencies

simu:
Kapitan has an option `force_fetch` to always fetch and update dependencies specified in `parameters.kapitan.dependencies`.

This commit sets that option to `True` to ensure component dependencies are always updated to their latest upstream variants.

This change removes the sometimes unwanted Kapitan behavior to keep stale dependencies, if the component either doesn't properly store dependencies in a version-specific directory, or if dependencies are specified with moving versions (e.g. `release-4.x` for OCP4 alert rule manifests).

## Checklist

### DIFF
--- a/commodore/helpers.py
+++ b/commodore/helpers.py
@@ -177,6 +177,7 @@ def kapitan_compile(
         cache=False,
         cache_paths=None,
         fetch_dependencies=fetch_dependencies,
+        force_fetch=True,
         validate=False,
         schemas_path=config.work_dir / "schemas",
         jinja2_filters=defaults.DEFAULT_JINJA2_FILTERS_PATH,


ccremer:
What is the implication to compiling with `--local` or `--no-fetch-dependencies` ?

simu:
* just `--local`: We already fetched Kapitan deps in "regular" local mode, with this commit we just ensure existing Kapitan dependency output files are overwritten
* `--local` and `--no-fetch-dependencies`: No dependencies are fetched, so the flag doesn't have an effect at all

ccremer (approved):

corvus-ch (approved):

PR #348, opened 2021-07-07
Correctly postprocess aliased components

simu:
Previously, we didn't respect component aliases when determining whether we need to run postprocessing.

## Checklist

- [x] Update tests.


corvus-ch (approved):
Renaming `inventory` to `inv` was a non functional change that made it harder to review. If your realy whant to do such types of changes, consider to do them in a PR on their own. Apart from that: LGTM.

PR #349, opened 2021-07-08
Add local dependencies search path in commodore test compile

bastjan:
See also / same as projectsyn/modulesync-control/pull/9.

Currently the `parameters.kapitan.dependencies.[].output_path` needs to be overridden for all dependencies in test compiles (`make test`).

See:
- https://github.com/projectsyn/component-cert-manager/blob/master/tests/defaults.yml#L9
- https://github.com/projectsyn/component-keycloak/blob/master/tests/builtin.yml#L10
- https://github.com/projectsyn/component-keycloak/blob/master/tests/external.yml#L10

This is not optimal. It can lead to errors if a developer forgets to update the `tests/` file. The "real" output can now differ from the test output. Additionally it won't scale for components like [`component-argocd`](https://github.com/projectsyn/component-argocd/blob/master/class/argocd.yml) where there are many dependecies.

This PR adds the local directory `./dependencies` to the search path, which allows tests to run without overriding the output path.

<details>
  <summary>Example: `component-cert-manager`</summary>
  
```diff
diff --git a/Makefile b/Makefile
index 59972dd..b76fb8e 100644
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ docs-serve: ## Preview the documentation
 	$(COMMODORE_CMD)
 
 .PHONY: test
-test: commodore_args = -f tests/$(instance).yml
+test: commodore_args = -f tests/$(instance).yml --search-paths ./dependencies
 test: .compile ## Compile the component
 
 .PHONY: clean
diff --git a/tests/defaults.yml b/tests/defaults.yml
index 8971146..739347e 100644
--- a/tests/defaults.yml
+++ b/tests/defaults.yml
@@ -1,12 +1,3 @@
 parameters:
-  kapitan:
-    dependencies:
-      - type: helm
-        source: https://charts.jetstack.io
-        chart_name: cert-manager
-        version: ${cert_manager:charts:cert-manager}
-        # This is the required overrride
-        output_path: /cert-manager/helmcharts/cert-manager/${cert_manager:charts:cert-manager}/
-
   cert_manager:
     letsencrypt_email: test@syn.tools
```
</details>

## Checklist

### DIFF
--- a/commodore/component-template/{{ cookiecutter.slug }}/Makefile
+++ b/commodore/component-template/{{ cookiecutter.slug }}/Makefile
@@ -42,6 +42,7 @@ docs-serve: ## Preview the documentation
 
 .PHONY: compile
 .compile:
+	mkdir -p dependencies
 	$(COMMODORE_CMD)
 
 .PHONY: test
--- a/commodore/component-template/{{ cookiecutter.slug }}/Makefile
+++ b/commodore/component-template/{{ cookiecutter.slug }}/Makefile
@@ -45,7 +45,7 @@ docs-serve: ## Preview the documentation
 	$(COMMODORE_CMD)
 
 .PHONY: test
-test: commodore_args = -f tests/$(instance).yml
+test: commodore_args = -f tests/$(instance).yml --search-paths ./dependencies
 test: .compile ## Compile the component
 
 .PHONY: clean


simu (approved):

ccremer (commented):
Is this really the only change required? It doesn't need to get volume-mounted from docker?
There are components that perform unit tests based on the output manifests

bastjan:
> Is this really the only change required? It doesn't need to get volume-mounted from docker?
> There are components that perform unit tests based on the output manifests after compilation

The parent directory is volume-mounted. Tested on a fresh checkout with the keycloak component which has go tests:

```sh
mkdir tmp-checkout && cd $_
git clone git@github.com:projectsyn/component-keycloak.git && cd component-keycloak
yq eval 'del(.parameters.kapitan)' -i tests/external.yml
yq eval 'del(.parameters.kapitan)' -i tests/builtin.yml
make test -e instance=builtin
make test -e instance=external
```

Works, but you were right, I needed to create the `./dependencies` directory before executing commodore.

simu:
> There are components that perform unit tests based on the output manifests after compilation

Dependencies are not output manifests. They're inputs to Kapitan, usually either Helm charts or YAMLs that we want to process in Jsonnet.

commodore/component-template/{{ cookiecutter.slug }}/Makefile
```diff
@@ -42,10 +42,11 @@ docs-serve: ## Preview the documentation
 
 .PHONY: compile
 .compile:
+	mkdir -p dependencies
```
simu, line 45: Doesn't seem required according to my local tests (Linux), can you elaborate why you think this is needed?

simu (approved):

commodore/component-template/{{ cookiecutter.slug }}/Makefile
bastjan: On "my machine"™️ (macos) commodore complains
```sh
$ docker run --rm -u "$(id -u)" -w /keycloak -v "${PWD}:/keycloak" projectsyn/commodore:latest component compile . -f tests/builtin.yml --search-paths ./dependencies
Error: Invalid value for '-J' / '--search-paths': Directory './dependencies' does not exist.
$ docker run --rm -u "$(id -u)" -w /keycloak -v "${PWD}:/keycloak" projectsyn/commodore:latest --version
commodore, version 0.8.2 (Git version: v0.8.2-6-g1788a4a)
```

simu: Ah yes, I was not able to properly test it, I get the same issue when the directory really doesn't exist :see_no_evil:

PR #351, opened 2021-07-28
Fix `render_target` to ensure instantiated components always compile

simu:
Previously, an error could occur when compiling a catalog containing an instantiated component for which no configurations were adjusted.

This commit ensures that the instance's key in `parameters` always exists.

## Checklist


simu:
I'm currently considering how a unit test for this case could look like, I didn't have a clear idea how to approach this yesterday.

corvus-ch (approved):

PR #352, opened 2021-07-29
List components with version used in a catalog

laserb:
List components with url and version that are used in a cluster catalog.

## Checklist

### DIFF
--- a/commodore/catalog.py
+++ b/commodore/catalog.py
@@ -1,12 +1,17 @@
 from pathlib import Path as P
-from typing import Iterable
+from typing import Iterable, Optional
 
 import click
 
 from . import git
-from .helpers import rm_tree_contents, lieutenant_query
-from .cluster import Cluster
+from .helpers import rm_tree_contents, lieutenant_query, ApiError, yaml_dump
+from .cluster import Cluster, update_target, update_params, load_cluster_from_api
 from .config import Config
+from .dependency_mgmt import _discover_components, _read_components
+from .fetch_config import (
+    fetch_global_config,
+    fetch_customer_config,
+)
 
 
 def fetch_customer_catalog(config: Config, cluster: Cluster):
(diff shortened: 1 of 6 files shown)


commodore/catalog.py
```diff
@@ -124,3 +129,42 @@ def catalog_list(cfg):
...
         else:
             click.echo(catalog_id)
+
+
+def list_components(config: Config, cluster_id: str, output: Optional[str] = None):
+    """
+    List all components required by target. Generate list of components
+    by searching for classes with prefix `components.` in the inventory files.
+
+    Component repos are searched in `GLOBAL_GIT_BASE/commodore_components`.
```
simu, line 139: This doesn't make any sense, `GLOBAL_GIT_BASE` doesn't exist anymore.

commodore/catalog.py
```diff
@@ -124,3 +129,42 @@ def catalog_list(cfg):
             click.echo(f" - {display_name}")
         else:
             click.echo(catalog_id)
+
+
+def list_components(config: Config, cluster_id: str, output: Optional[str] = None):
+    """
+    List all components required by target. Generate list of components
+    by searching for classes with prefix `components.` in the inventory files.
```
simu, line 137: I assume you've copied the function comment from `fetch_components`. I just noticed that that comment is outdated.

Since #227, the list of components included for the cluster is extracted from the `applications` array as described in https://syn.tools/commodore/reference/architecture.html#_component_discovery_and_versions

commodore/compile.py
```diff
@@ -67,8 +41,8 @@ def _regular_setup(config: Config, cluster_id):
     update_params(config.inventory, cluster)
 
     # Fetch components and config
-    _fetch_global_config(config, cluster)
-    _fetch_customer_config(config, cluster)
+    fetch_global_config(config, cluster)
+    fetch_customer_config(config, cluster)
```
simu, line 45: Might be worth to consider refactoring everything in `_regular_setup` up to this line into another method which can be used here and for `list_components`.

commodore/catalog.py
```diff
@@ -124,3 +129,42 @@ def catalog_list(cfg):
             click.echo(f" - {display_name}")
         else:
             click.echo(catalog_id)
+
+
+def list_components(config: Config, cluster_id: str, output: Optional[str] = None):
```
simu, line 134: This method doesn't really belong in `catalog.py`, since it doesn't interact with the cluster catalog at all. If I had to place it somewhere, I'd probably put it in `cluster.py`.

commodore/catalog.py
```diff
@@ -124,3 +129,42 @@ def catalog_list(cfg):
...
+        update_target(config, config.inventory.bootstrap_target)
+        update_params(config.inventory, cluster)
+
+        # Fetch components and config
+        fetch_global_config(config, cluster)
+        fetch_customer_config(config, cluster)
+
+        click.secho("Discovering components...", bold=True)
+        component_names, component_aliases = _discover_components(config)
+        urls, versions = _read_components(config, component_names)
```
simu, line 162: If you use "private" methods outside the module, please change them to not be "private" anymore.

commodore/helpers.py
```diff
@@ -73,13 +74,16 @@ def _represent_str(dumper, data):
...
     Dump obj as single-document YAML
     """
     yaml.add_representer(str, _represent_str)
-    with open(file, "w") as outf:
-        yaml.dump(obj, outf)
+    if file is None:
+        yaml.dump(obj, sys.stdout)
+    else:
+        with open(file, "w") as outf:
+            yaml.dump(obj, outf)
```
simu, line 86: This seems unnecessary given the latest version, as you're always passing a value for `file`.

simu (changes requested):
Please also make sure to update the documentation (at least `docs/modules/ROOT/pages/reference/cli.adoc` since you're adding a new CLI command).

Additionally, it would be nice to have a test for `list_components`, although it's not really critical since most of the building blocks for the function are already tested.

(there's a reason this checkbox is in the PR template)

commodore/helpers.py
laserb: The file value is from an optional command line option, which can be undefined/None.

simu:
Closing this, as I don't see the need for it at the moment, and my comments haven't been addressed.

PR #353, opened 2021-08-04
Fix postprocessing filter lookup

simu:
This commit makes use of the information provided in Component objects to do the postprocessing filter lookup for filters of type `jsonnet`.

This fixes the "Jsonnet filter definition does not exist" error which previously popped up when running trying to compile components with postprocessing filters with `commodore component compile .`.

## Checklist

### DIFF
--- a/commodore/postprocess/jsonnet.py
+++ b/commodore/postprocess/jsonnet.py
@@ -8,6 +8,7 @@ from typing import Any, Callable, Dict, Iterable
 import _jsonnet
 
 from commodore.config import Config
+from commodore.component import Component
 from commodore.helpers import yaml_load, yaml_load_all, yaml_dump, yaml_dump_all
 from commodore import __install_dir__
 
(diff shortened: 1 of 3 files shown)


commodore/postprocess/jsonnet.py
```diff
@@ -115,15 +116,14 @@ def _inventory() -> Dict[str, Any]:
             yaml_dump(outcontents, outpath)
 
 
-def _filter_file(work_dir: P, component: str, filterpath: str) -> P:
-    # TODO: Do we need to handle search path better?
-    return work_dir / "dependencies" / component / filterpath
+def _filter_file(component: Component, filterpath: str) -> P:
+    return component.target_directory / filterpath
```
simu, line 120: The PR diff is a bit noisy because I had to change the whole postprocessing infrastructure to work with the `Component` object instead of the component name, so that this change is possible.

corvus-ch (approved):

ccremer (approved):

PR #355, opened 2021-08-04
Implement optional `output_dir` postprocess filter argument

simu:
Previously Commodore postprocessing filters had the hard-coded assumption that each component's output would appear in `compiled/<component-name>`. This is not necessarily true, e.g. when an instance-aware component creates individual ArgoCD apps and folders for each instance.

This PR introduces an optional parameter `output_dir` for postprocessing filter definitions which allows users to override the default directory in `compiled` in which Commodore applies postprocessing filters for the component.

## Checklist

- [x] Rebase after #353 has been merged
- [x] Update tests.

### DIFF
--- a/commodore/postprocess/builtin_filters.py
+++ b/commodore/postprocess/builtin_filters.py
@@ -13,8 +13,9 @@ from commodore.component import Component
 from .jsonnet import jsonnet_runner
 
 
-def _output_dir(work_dir: P, component: str, path):
-    return work_dir / "compiled" / component / path
+def _output_dir(work_dir: P, compiled_dir: str, path):
+    """Compute directory in which to apply filter"""
+    return work_dir / "compiled" / compiled_dir / path
 
 
 def _builtin_filter_helm_namespace(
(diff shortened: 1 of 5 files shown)


commodore/postprocess/builtin_filters.py
```diff
@@ -31,7 +31,9 @@ def _builtin_filter_helm_namespace(
         create_namespace = "true" if create_namespace else "false"
     exclude_objects = kwargs.get("exclude_objects", [])
     exclude_objects = "|".join([json.dumps(e) for e in exclude_objects])
-    output_dir = _output_dir(work_dir, component.name, path)
+    # XXX: we pass "" for the component name here, as we already patch
+    # `path` to contain the output dir in postprocess/__init__.py
```
ccremer, line 35: What's with the XXX?

ccremer (approved):

commodore/postprocess/builtin_filters.py
simu: Changed to `NOTE:` in the latest version

commodore/postprocess/builtin_filters.py
```diff
@@ -31,7 +32,9 @@ def _builtin_filter_helm_namespace(
         create_namespace = "true" if create_namespace else "false"
     exclude_objects = kwargs.get("exclude_objects", [])
     exclude_objects = "|".join([json.dumps(e) for e in exclude_objects])
-    output_dir = _output_dir(work_dir, component.name, path)
+    # NOTE: we pass "" for compiled_dir here, as we already patch `path` to
```
corvus-ch, line 35: Sounds like a redundancy that could be eliminated.

corvus-ch (approved):
There might be room for improvement but I do not have hard fealings on that one.

commodore/postprocess/builtin_filters.py
simu: yeah, I looked at it again, and figured out that I implemented a workaround for a bug in the postprocessing in this PR :see_no_evil: 

See #357 for the correct fix.

PR #363, opened 2021-08-25
Make yamllint exludes work on both BSD and GNU

bastjan:
`find` works differently on GNU and BSD.

BSD find would need an -E option or have the backslash before the braces removed.

## Checklist


glrf (approved):
I mean I don't expect anyone to develop on BSD, but using `.yamllint.yml` seems cleaner anyway.

PR #364, opened 2021-08-26
Add golden and matrix tests to component template

bastjan:
See https://github.com/projectsyn/modulesync-control/pull/22 as to why golden tests can be a good idea.

## Checklist

### DIFF
--- a/commodore/component-template/{{ cookiecutter.slug }}/.gitignore
+++ b/commodore/component-template/{{ cookiecutter.slug }}/.gitignore
@@ -1,7 +1,18 @@
-_archive/
-_public/
+# Commodore
 .cache/
 helmcharts/
 manifests/
 vendor/
 jsonnetfile.lock.json
+crds/
+compiled/
+
+# Editors
+.vscode/
+.idea/
+
+# Antora
+_archive/
+_public/
+
+# Additional entries
--- a/commodore/component-template/{{ cookiecutter.slug }}/.gitignore
+++ b/commodore/component-template/{{ cookiecutter.slug }}/.gitignore
@@ -7,10 +7,6 @@ jsonnetfile.lock.json
 crds/
 compiled/
 
-# Editors
-.vscode/
-.idea/
-
 # Antora
 _archive/
 _public/
--- a/commodore/component-template/{{ cookiecutter.slug }}/Makefile
+++ b/commodore/component-template/{{ cookiecutter.slug }}/Makefile
@@ -49,6 +49,7 @@ docs-serve: ## Preview the documentation
 test: commodore_args += -f tests/$(instance).yml
 test: .compile ## Compile the component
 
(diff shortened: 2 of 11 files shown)


commodore/component-template/{{ cookiecutter.slug }}/.gitignore
```diff
@@ -1,7 +1,18 @@
...
+# Commodore
 .cache/
 helmcharts/
 manifests/
 vendor/
 jsonnetfile.lock.json
+crds/
+compiled/
+
+# Editors
```
corvus-ch, line 10: This is subject to grow out of hand. We do not have an agreement on which editor to use. In consequence, every editor can be used. Adding ignores for every editor is nothing I would be willing to do.

As a git user, I consider it to be best practice, to have a global git ignore to exclude things specific to my system. This include also such things like vim swap files, the all famous and anoying `.DS_Store` macOS etc.

commodore/component-template/{{ cookiecutter.slug }}/Makefile
```diff
@@ -49,6 +49,18 @@ docs-serve: ## Preview the documentation
 test: commodore_args += -f tests/$(instance).yml
 test: .compile ## Compile the component
 
+.PHONY: gen-golden
```
corvus-ch, line 52: Do we want to have the targets even if it is not enabled for the component? Consider to wrap them in conditions.

corvus-ch (changes requested):
Please do not open up the think hole of adding editor specific git ignores.

commodore/component-template/{{ cookiecutter.slug }}/.gitignore
simu: I agree that editor-specific things don't belong in repo .gitignores but rather in a global .gitignore or (`.git/info/exclude`). However, the contents of this gitignore are verbatim from https://github.com/projectsyn/modulesync-control/blob/master/moduleroot/.gitignore.erb, so if we want to change it here, we need to change it there too.

commodore/component-template/{{ cookiecutter.slug }}/Makefile
simu: I totally missed those when implementing the flags

commodore/component-template/{{ cookiecutter.slug }}/.gitignore
simu: Updated in latest version and removed from modulesync-control: https://github.com/projectsyn/modulesync-control/commit/668a2b3edfef9e129100a8aecda1dd08f9a891bf

commodore/component-template/{{ cookiecutter.slug }}/Makefile
simu: added in latest version

corvus-ch (approved):

PR #379, opened 2021-11-08
Allow ARM Docker build

bastjan:
## Checklist

### DIFF
--- a/tools/install-jb.sh
+++ b/tools/install-jb.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+set -eu
+
+VERSION=${1:-}
+
+if [ "x${VERSION}" = "x" ]; then
+	echo "Usage: $0 VERSION"
+	exit 3
+fi
+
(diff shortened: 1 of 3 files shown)


tools/install-jb.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+set -eu
+
+VERSION=${1:-}
+
+if [ "x√" = "x" ]; then
```
simu, line 6: This comparison looks pretty broken, are you sure you didn't want

```suggestion
if [ "x${VERSION}" = "x" ]; then
```

simu (approved):
LGTM overall.

Side note (although I'm sure you're aware of that): this PR doesn't actually build and publish ARM images on DockerHub.

tools/install-jb.sh
bastjan: Yep that is super broken 🙈

PR #383, opened 2021-11-12
Add CLI command to extract component versions and URLs

simu:
This PR introduces a new command group called `inventory` and a new command in that group called `inventory components`.

The new command `inventory components` calls reclass to render an already checked-out Commodore defaults repo and returns all components (URLs and versions) defined in that repo. The command has a repeatable optional argument `-f / --values` to specify additional inventory classes in YAML files.

Users should use `-f` to provide additional facts (such as the cluster's K8s distribution) to the command as well as providing default values for parameters which are used in the global defaults repo without a configured default.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/inventory/parameters.py
+++ b/commodore/inventory/parameters.py
@@ -1,11 +1,12 @@
 import os
 from enum import Enum
 from pathlib import Path
-from typing import Any, Dict, Iterable, Optional
+from typing import Any, Dict, Iterable, List, Optional
 
 from kapitan.reclass import reclass
 from kapitan.reclass.reclass import settings as reclass_settings
 from kapitan.reclass.reclass import core as reclass_core
+from kapitan.reclass.reclass import errors as reclass_errors
 
 from commodore.cluster import Cluster, render_params
 from commodore.component import component_parameters_key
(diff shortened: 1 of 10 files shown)


commodore/inventory/parameters.py
```diff
@@ -0,0 +1,229 @@
...
+                "id": "t-foo",
+                "displayName": "Foo Inc.",
+                "gitRepo": {
+                    "url": "not-a-real-repo",
+                },
+            },
+        )
+        params = render_params(self._inventory, cluster)
+        # don't support legacy component_versions key
+        del params["parameters"]["components"]
```
glrf, line 162: I'm not quite sure why you remove this here. Probably makes sense, but I'm missing the context.

commodore/inventory/parameters.py
```diff
@@ -0,0 +1,229 @@
...
+        del params["parameters"]["component_versions"]
+        params["parameters"]["openshift"] = {
+            "infraID": "infra-id",
+            "clusterID": "cluster-id",
+            "registryBucket": "x-fake-registry-bucket",
+        }
+        params["parameters"]["cloud"]["availabilityZones"] = []
+        params["parameters"]["cloud"]["projectName"] = "x-fake-project"
```
glrf, line 170: I guess these are the VSHN specific parameter you mentioned?

glrf (approved):
I tried to understand how this works and ran it locally and I did not find any obvious issues so LGTM.

commodore/inventory/parameters.py
```diff
@@ -0,0 +1,229 @@
...
+                "distribution": distribution,
+                "cloud": cloud,
+                "lieutenant-instance": "lieutenant-prod",
+                f"{distribution}_version": "1.20",
+            },
+            "gitRepo": {
+                "url": "not-a-real-repo",
+            },
+        }
+        if region:
```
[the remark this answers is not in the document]
corvus-ch: I second this.

commodore/inventory/parameters.py
```diff
@@ -0,0 +1,229 @@
...
+        if cloud is None:
+            cloud = "x-fake-cloud"
+        if region is None:
+            region = "x-fake-region"
+
+        c: Dict[str, Any] = {
+            "id": "c-bar",
+            "tenant": "t-foo",
+            "displayName": "Foo Inc. Bar cluster",
+            "facts": {
```
corvus-ch, line 137: I assume that we will add more and more dynamic facts. Having to touch this each time we do so, would be a nuisance. Instead of handling the known facts explicitly, I suggest to do this in a more generic way.

docs/modules/ROOT/pages/reference/cli.adoc
```diff
@@ -152,3 +152,17 @@ This command doesn't have any command line options.
 
 *--help*::
   Show component new usage and options then exit.
+
+== Inventory Components
+
+*-d, --distribution*::
```
corvus-ch, line 158: I suggest to go with a more generic approach. Can we add parameters that can be given more than once? If yes, can we have just a parameter to provide facts and one to provide parameters. With this, we then could also remove setting defaults for `parameters.cloud.availabilityZones` and `parameters.cloud.projectName`. Yes I am aware that this would make usage for as a bit more involved (adding them with each and every call) but it would keep this code a bit more clutter free. Alternatively, allow to pass them as a path to a yaml file.

corvus-ch (changes requested):
Proably I no longer recall how deep some facts are anchored into the Commodore code base. However, I would prefer a more generic apraoch. Instead of having a given set, have a way to provide arbitrary facts and paramters to be merged into reclass.

commodore/inventory/parameters.py
simu: Refactored the command in the latest version to not need those anymore.

commodore/inventory/parameters.py
simu: It's probably not strictly necessary here, but this ensures we don't consider component versions or URLs which are still configured in `parameters.component_versions` (cf. https://syn.tools/commodore/reference/deprecation-notices.html#_v0_5_0)

docs/modules/ROOT/pages/reference/cli.adoc
simu: I've implemented the variant which allows passing an arbitrary number of extra YAML files which are used as additional inventory classes.

corvus-ch (approved):

PR #390, opened 2021-12-06
Add deprecation notice for component specifications without explicit version

### LINKED ISSUE #374: Deprecate component specifications without an explicit version
## Context

We introduced explicit versioning as the new recommended default for Commodore components in #284. However, we never followed up and actually made component specifications without explicit versions deprecated.

In order to be able to remove support for component specifications without explicit versions in Commodore 1.0 we need to deprecate such specifications in a final 0.x release.

## Alternatives

n/a

simu:
Fixes #374

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/dependency_mgmt.py
+++ b/commodore/dependency_mgmt.py
@@ -95,8 +95,8 @@ def _read_components(
             component_versions[component_name] = info["version"]
         else:
             cfg.register_deprecation_notice(
-                f"Component {component_name} doesn't have a version specified."
-                + "See https://syn.tools/commodore/reference/deprecation-notices.html#_components_without_versions for more details"
+                f"Component {component_name} doesn't have a version specified. "
+                + "See https://syn.tools/commodore/reference/deprecation-notices.html#_components_without_versions for more details."
             )
             # Note: We use version=None as a marker for checking out the remote repo's
             # default branch.
--- a/commodore/dependency_mgmt.py
+++ b/commodore/dependency_mgmt.py
@@ -94,7 +94,12 @@ def _read_components(
         if "version" in info:
             component_versions[component_name] = info["version"]
         else:
-            # Note: We use version=None as a marker for checking out the remote repo's default branch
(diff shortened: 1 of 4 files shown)


commodore/dependency_mgmt.py
```diff
@@ -95,7 +95,12 @@ def _read_components(
         if "version" in info:
             component_versions[component_name] = info["version"]
         else:
-            # Note: We use version=None as a marker for checking out the remote repo's default branch
+            cfg.register_deprecation_notice(
+                f"Component {component_name} doesn't have a version specified."
+                + "See https://syn.tools/commodore/reference/deprecation-notices.html#_components_without_versions for more details"
```
ccremer, line 100: I'm no Python expert, but isn't there a newline or whitespace missing between the 2 sentences or does Python do this automatically?

ccremer (approved):

commodore/dependency_mgmt.py
simu: It was missing, fixed in the latest version

tests/test_catalog_compile.py
```diff
@@ -211,6 +211,10 @@ def test_catalog_compile(load_cluster, config: Config, tmp_path: Path, capsys):
     # Stdout success msg
     captured = capsys.readouterr()
     assert "Catalog compiled!" in captured.out
+    assert (
```
corvus-ch, line 214: Can we have a case that tests the deprecation notice is shown?

corvus-ch (approved):
Maybe add another test test for when the notice is shown. Apart from that, everything looks fine to me.

tests/test_catalog_compile.py
simu: I didn't extend the integration test, since that requires a lot of plumbing to test multiple different configurations (another catalog repo on GitHub etc.), but I added unit tests for `dependency_mgmt._read_components()`.

PR #392, opened 2021-12-06
Add support for tenant repo in `commodore inventory components`

simu:
Will be used in https://github.com/projectsyn/commodore-renovate/pull/5

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/inventory/parameters.py
+++ b/commodore/inventory/parameters.py
@@ -10,8 +10,11 @@ from kapitan.reclass.reclass import errors as reclass_errors
 
 from commodore.cluster import Cluster, render_params
 from commodore.component import component_parameters_key
-from commodore.helpers import yaml_dump
+from commodore.helpers import yaml_dump, yaml_load
 from commodore.inventory import Inventory
+
+FAKE_TENANT_ID = "t-foo"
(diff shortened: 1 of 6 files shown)


commodore/inventory/parameters.py
```diff
@@ -259,17 +293,31 @@ def cloud_regions(self) -> Dict[str, Iterable[str]]:
         return self._cloud_regions
 
     @classmethod
-    def _make_directories(cls, work_dir: Path):
+    def _make_directories(cls, work_dir: Path, fake_tenant=True):
```
glrf, line 296: Very much personal preference, but IMO this helper function does nothing to improve readability. It's used once and `from_repo_dirs` needs to work with the classes directory anyway

glrf (approved):
LGTM as far as I can tell. One very minor nitpick that I would have done differently, but feel free to ignore that :)

commodore/inventory/parameters.py
simu: I think I added this in #383 because I thought it might be useful when implementing the functionality for cloning global or tenant repos in the command (which still isn't there).

I've inlined it for now.

corvus-ch (approved):

PR #396, opened 2021-12-28
Use absolute Docker Image URLs

mhutter:
Especially on Podman systems, docker.io may not be the default, thus breaking image pulls.

Be explicit where to pull images from by configuring a registry.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,8 @@ MAKEFLAGS += -j4
 docker_cmd  ?= docker
 docker_opts ?= --rm --tty --user "$$(id -u)"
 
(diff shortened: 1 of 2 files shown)


simu (approved):
The modulesync-control template (https://github.com/projectsyn/modulesync-control/blob/master/moduleroot/Makefile.vars.mk.erb) should be updated as well, if the component template in Commodore is changed, otherwise modulesync will revert the absolute image URLs in the component template when components are onboarded to modulesync.

mhutter:
Good point, projectsyn/modulesync-control#53

PR #400, opened 2022-01-19
Add new command `inventory lint`

simu:
Currently the only implemented linting error is warning about component specifications without explicit version.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/cli.py
+++ b/commodore/cli.py
@@ -1,6 +1,7 @@
 import json
+import sys
 from pathlib import Path
-from typing import Iterable, Optional
+from typing import Iterable, Optional, Tuple
 
 import click
 import yaml
@@ -15,6 +16,7 @@ from .component.template import ComponentTemplater
 from .component.compile import compile_component
 from .inventory.render import extract_components
 from .inventory.parameters import InventoryFacts
+from .inventory.lint_components import lint_components
 
 pass_config = click.make_pass_decorator(Config)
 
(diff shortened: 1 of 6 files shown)


commodore/cli.py
```diff
@@ -417,6 +419,46 @@ def component_versions(
         click.echo(yaml.safe_dump(components))
 
 
+@inventory.group(
+    name="lint", short_help="Commands which lint the Commodore inventory structure"
+)
+@verbosity
+@pass_config
+def inventory_lint(config: Config, verbose):
+    config.update_verbosity(verbose)
```
glrf, line 428: Could we make this command lint everything (i.e. execute all it's sub commands)?

I pretty much never want to just check for one type of issues. I would want commodore to just lint my inventory. I don't want to run a command to check if there are components without versions, then another one check for a different issue, etc.

glrf (approved):
LGTM. 
I would have structured the CLI differently, but that's probably personal preference,

commodore/cli.py
simu: Not sure if that's possible with the current structure, but we can certainly refactor the CLI to just provide `inventory lint` and then options to enable or disable individual linters. Thinking about it again, that probably makes more sense than my current approach.

corvus-ch: Where would that linting be executed?

corvus-ch (commented):
I fail to understand where that linting would be executed.

simu:
> I fail to understand where that linting would be executed.

Anywhere you can run Commodore really. This can be useful locally to check that you didn't introduce an unsupported configuration, or in CI for a global or tenant repo to catch unsupported configs before they get merged to the main branch.

commodore/cli.py
simu: Changed the CLI to just provide `inventory lint`. We can always add flags to enable or disable individual linters once we have more than one.

commodore/cli.py
```diff
@@ -417,6 +419,43 @@ def component_versions(
...
+    name="lint",
+    short_help="Lint YAML files for Commodore inventory structures in the provided paths",
+    no_args_is_help=True,
+)
+@click.argument(
+    "target", type=click.Path(file_okay=True, dir_okay=True, exists=True), nargs=-1
+)
+@verbosity
+@pass_config
+def inventory_lint(config: Config, verbose: int, target: Tuple[str], components: bool):
```
glrf, line 432: I've never used click, but don't we need to specify `components` as an option?

tests/test_cli.py
```diff
@@ -55,3 +55,8 @@ def test_component_compile_command():
 def test_component_inventory_components_command():
     exit_status = call("commodore inventory components --help", shell=True)
     assert exit_status == 0
+
+
+def test_component_inventory_lint_components_command():
+    exit_status = call("commodore inventory lint components --help", shell=True)
+    assert exit_status == 0
```
glrf, line 62: That command doesn't exist anymore?

simu: Yep, missed that one too. The test still passes because `commodore inventory lint components --help` gets shortened to `commodore inventory lint --help` by Click.

PR #403, opened 2022-01-21
Cache Python virtualenvs for repeated CI runs with the same dependencies

simu:
## Checklist

- [ ] Update tests.
- [ ] Categorize the PR by setting a good title and adding one of the labels:


simu:
The last CI run didn't reuse the cache, because `actions/cache@v2` internally also uses the configured paths as part of the cache key, cf. https://github.com/actions/cache/blob/611465405cc89ab98caca2c42504d9289fb5f22e/src/restore.ts#L36-L40

simu:
See https://github.com/projectsyn/commodore/actions/runs/1729326932 for a CI run which successfully used the cached virtualenvs.

glrf (approved):

PR #406, opened 2022-01-28
Add GitHub action to build virtualenv caches on master branch 

simu:
## Checklist


simu:
See https://github.com/projectsyn/commodore/runs/4982762261?check_suite_focus=true for an example where the cache built on the master branch has been reused already for this PR

PR #430, opened 2022-03-25
Add `kustomize` to container image

bastjan:
Adds `kustomize` as a usable tool to the Commodore Docker image.
Script taken from upstream.

I'll write up a documentation how to use it at a later point after some experience.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,12 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master
  && ./get_helm.sh \
  && mv /usr/local/bin/helm /usr/local/bin/helm2
 
-RUN ./tools/install-jb.sh v0.4.0
+ARG KUSTOMIZE_VERSION=4.5.7
+
+RUN ./tools/install-jb.sh v0.4.0 \
+ && curl -fsSLO "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" \
+ && chmod +x install_kustomize.sh \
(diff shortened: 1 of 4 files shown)


tools/install-kustomize.sh
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env bash
+
+## https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh
```
simu, line 3: If this is a 1:1 copy of the upstream script, I'd suggest using `curl` to fetch the script in the `Dockerfile` instead of keeping a copy in `tools/`.

Dockerfile
```diff
@@ -41,7 +41,8 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master
  && ./get_helm.sh \
  && mv /usr/local/bin/helm /usr/local/bin/helm2
 
-RUN ./tools/install-jb.sh v0.4.0
+RUN ./tools/install-jb.sh v0.4.0 \
+ && ./tools/install-kustomize.sh 4.5.7 /usr/local/bin
```
simu, line 45: We may want to consider configuring Renovate for the Kustomize version (probably just with regex manager).

simu (changes requested):
Re documentation: we should at least add Kustomize to the prerequisites in `explanation/running-commodore.adoc`.

Dockerfile
bastjan: Added a regex manager.

simu (approved):

PR #431, opened 2022-03-28
Auto-login when trying to authenticate without token

glrf:
## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/config.py
+++ b/commodore/config.py
@@ -1,8 +1,11 @@
+import time
 import textwrap
 
 from enum import Enum
 from pathlib import Path as P
 from typing import Dict, List, Optional
+
+import jwt
 
 import click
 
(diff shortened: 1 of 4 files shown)


commodore/config.py
```diff
@@ -104,7 +107,19 @@ def refs_dir(self) -> P:
...
-            self._api_token = tokencache.get(self.api_url)
+            token = tokencache.get(self.api_url)
+            if token is not None:
+                # We don't verify the signature, we just want to know if the token is expired
+                # lieutenant will decide if it's valid
+                try:
+                    t = jwt.decode(
+                        token, algorithms=["RS256"], options={"verify_signature": False}
+                    )
+                    if "exp" in t and t["exp"] < time.time():
```
simu, line 118: This comparison leaves a slight chance of ending up trying to use an expired token if the token expires in a few seconds, and we perform the actual API calls after it's expired.

Maybe better:

```suggestion
                    if "exp" in t and t["exp"] < time.time() + 10:
```

simu (approved):

PR #432, opened 2022-03-29
Update component template to support Podman out of the box

mhutter:
While it was already possible to use podman by setting the DOCKER_CMD
env var, this was not enough for most commands.

This commit adds support for Podman out of the box. If Docker is
installed, it will be preferred over Podman to stay
backwards-compatible.

The `--userns=keep-id` flag is required on Podman to ensure that 1.) the
process in the container can write to the mounted workdirs, and 2.)
ensure files created by processes in the container have the correct owner
and group IDs: The ones from the user running the Make tasks.

## Checklist

- [x] Update tests.


simu (approved):
Strictly speaking this should probably be 2 PRs:

1. "Update component template to support Podman out of the box" (`enhancement`)
2. "Update Commodore Makefile to support Podman out of the box" (`change`)

PR #441, opened 2022-04-01
Move test coverage to own GitHub action

simu:
This allows us to run only the coverage test run on pushes to master without having to resort to conditional jobs for all other tests.

We need to run the coverage report on pushes to master to actually get visible code coverage reports from codeclimate.

Follow-up for #439 

## Checklist

- [x] Update tests.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,51 @@
+name: Test coverage with codeclimate
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+
+jobs:
+  coverage:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v3
+      with:
+        python-version: '3.9'
+    - uses: Gr1N/setup-poetry@v7
+    - uses: actions/cache@v3
+      with:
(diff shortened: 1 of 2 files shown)


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,51 @@
...
+        poetry env use python3.9
+        poetry install
+    - name: Install jsonnet-bundler
+      run: |
+        mkdir -p /opt/bin && curl -sLo /opt/bin/jb \
+          https://github.com/jsonnet-bundler/jsonnet-bundler/releases/download/v0.4.0/jb-linux-amd64 \
+          && chmod +x /opt/bin/jb
+    - name: Update PATH
+      run: echo "/opt/bin" >> $GITHUB_PATH
+    - name: Pull in SSH deploy key for integration test
```
corvus-ch, line 34: What do we need the deployment key for?

corvus-ch (approved):
I do not get why coverage reporting needs access to the deployment key. Apart from that, LGTM.

simu:
> I do not get why coverage reporting needs access to the deployment key. Apart from that, LGTM.

We run the integration test as part of the coverage action (cf. https://github.com/projectsyn/commodore/commit/fb07390fd35c946a7557559af986031e9edb6a8d), which needs the deploy key for the integration cluster catalog.

PR #450, opened 2022-04-04
Refactor `_push_catalog()` to return early in local mode

simu:
## Checklist

- [x] Update tests

### DIFF
--- a/tests/test_catalog.py
+++ b/tests/test_catalog.py
@@ -3,7 +3,9 @@ Tests for catalog internals
 """
 import copy
 from pathlib import Path
+from unittest.mock import patch
 
+import click
 import git
 import pytest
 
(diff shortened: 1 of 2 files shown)


tests/test_catalog.py
```diff
@@ -106,3 +108,128 @@ def test_fetch_catalog_initial_commit_for_empty_repo(
...
+    repo = catalog.fetch_catalog(config, cluster)
+    with open(repo.working_tree_dir / "test.txt", "w") as f:
+        f.writelines(["foo"])
+
+    repo.stage_files(["test.txt"])
+
+    return repo
+
+
+# Commodore doesn't support `catalog compile --push --local`, so we omit that test case.
```
corvus-ch, line 123: Is there a test to check this is indeed not possible?

corvus-ch (approved):

tests/test_catalog.py
simu: I realized that the logic preventing it is right there in `_push_catalog()` so I changed the test to check all combinations of `--local` and `--push`.

Issue #375, opened 2021-11-03
Remove all currently deprecated features before releasing v1.0


simu:
## Context

Over time we've deprecated quite a few features in Commodore, cf. https://syn.tools/commodore/reference/deprecation-notices.html. Before we release v1.0 we should remove all those deprecated features which are still present in the implementation.

## Alternatives

n/a

simu:
List of features to remove

- [ ] Version overrides in `parameters.component_versions`
- [ ] Component specifications without versions
- [ ] External postprocessing filter definitions
- [ ] Top-level multi-instance advertisement
- [ ] `parameters.cluster.dist`, `parameters.cloud.provider`, `parameters.cloud.region`, `parameters.customer.name`

PR #453, opened 2022-04-05
Remove support for external postprocessing filter definitions

simu:
We can completely remove `postprocess/inventory.py` since we now always get rendered inventory values from Reclass/Kapitan when reading filter definitions.

Also, we remove the `test_postprocess_components_enabledref` test, because that test only made sense for external postprocessing filters without actually rendering the mocked inventory with Reclass.

Relates #375 

## Checklist

- [x] Update tests.


corvus-ch (approved):
Yes, burn that old code but with fire 😛

PR #468, opened 2022-04-07
Normalize YAML

mhutter:
This commit tries to get rid of most yamllint warnings

- Some of the component template files were not compliant with the very
  yamllint config included therein (als see projectsyn/modulesync-control#73)
- Many files used different indentations in the same file

All in all this commit should make it easier to spot actual errors when
using yamllint.

## Checklist

- [ ] Update tests.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,3 +1,4 @@
+---
 version: "2"
 
 checks:
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,3 +1,4 @@
+---
 blank_issues_enabled: false
 contact_links:
   - name: ❓ Help and Support RocketChat Channel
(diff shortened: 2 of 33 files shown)


simu (commented):
> Some of the component template files were not compliant with the very yamllint config included therein.

Only if you enable warnings which the component template's `make lint_yaml` doesn't do by default. See also the [test_check_component_template](https://github.com/projectsyn/commodore/blob/5152cb230a899f90228455c6854fa78a5e473859/tests/test_component_template.py#L285-L307) tests which ensure that the template passes the lint.

If we don't want any warnings, we should also enable warnings in the `lint_yaml` make target of the component template.

Also, this PR should probably be split into two, one which only deals with the component template (that one could then be tagged `enhancement`), and one which makes the changes for the rest of the YAML files in the Commodore repository (probably with tag `change`).

simu:
Closing this PR since we've moved the component template out of the Commodore repository into its own repository. Issue to track the request on the template repository: https://github.com/projectsyn/commodore-component-template/issues/10

PR #472, opened 2022-04-07
Add integration test for `catalog compile --local`

simu:
## Checklist

- [x] Update tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ inject-version:
 
 .PHONY: test_integration
 test_integration:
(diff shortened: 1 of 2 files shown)


glrf (approved):

corvus-ch (approved):
I dislike all those path concatenations. I would prefer to use the properties/functions of the component class. Reasoning: building those paths is not the goal of this test. Recreating logic makes this test fragile to change.

Looking at the `Component` class, this does not seem to work, as the values used there are not available there.

PR #484, opened 2022-04-12
Convert type annotations for standard collections to PEP-585 

simu:
PEP-585 was introduced for Python 3.9. We use `from __future__ import annotations` for Python 3.8 compatibility.

See https://peps.python.org/pep-0585/ for details

## Checklist


simu:
> Just scrolled through, no super deep review

That's fine, this is a purely non-functional change.

PR #498, opened 2022-05-03
Refactor `commodore.dependency_mgmt` into a package

simu:
In preparation for implementing support for configuration packages (cf. #497), we refactor `commodore.dependency_mgmt` into a multi-file package.

This PR also moves frequently used pytest fixtures to `tests/conftest.py`. This allows us to use them in all tests without redefining them in each test file. Additionally, the commit normalizes the naming (having a fixture `data` is not particularly helpful when trying to figure out what it does).

Finally, we adjust the tox config to run tox in an empty temp dir. This ensures we don't get import conflicts, since pytest now changes the python import paths because it detects a `conftest.py`.

NOTE: This PR doesn't make any functional changes.

Factored out from #497 to allow independent review.

## Checklist

- [x] Keep pull requests small so they can be easily reviewed. (Bulk of the changes caused by moving code & renaming fixtures)


glrf (approved):
That's a huge diff, so I might have missed something, but the restructuring looks good to me

PR #505, opened 2022-05-12
Add command `commodore package compile`

simu:
## Checklist

- [x] Keep pull requests small so they can be easily reviewed.
- [x] Update the documentation.
- [x] Update tests.
- [x] Categorize the PR by setting a good title and adding one of the labels:
      `bug`, `enhancement`, `documentation`, `change`, `breaking`, `dependency`
      as they show up in the changelog
- [x] Link this PR to related issues.

PR #506, opened 2022-05-13
Add flags `--local` and `--no-fetch-dependencies` for `package compile`

### LINKED ISSUE #499: Implement new command to compile configuration packages standalone
## Context

In #497 we implement support for configuration packages in `commodore catalog compile`. However, the PR doesn't provide a new command to compile configuration packages standalone, for example to allow easy CI/CD for package repos.

A full implementation of the [configuration packages SDD](https://syn.tools/syn/SDDs/0028-reusable-config-packages.html) requires this command to be present.

## Alternatives

Leave the SDD half-implemented

simu:
Add command line flags `--local` and `--no-fetch-dependencies` with the same semantics as for `catalog compile`.

Follow-up for #505 
Resolves #499

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/package/compile.py
+++ b/commodore/package/compile.py
@@ -10,7 +10,7 @@ from commodore.compile import clean_working_tree
 from commodore.cluster import update_target
 from commodore.compile import setup_compile_environment
 from commodore.config import Config
-from commodore.dependency_mgmt import fetch_components
+from commodore.dependency_mgmt import fetch_components, register_components
 from commodore.helpers import kapitan_compile, relsymlink, yaml_dump
 from commodore.inventory import Inventory
 from commodore.postprocess import postprocess_components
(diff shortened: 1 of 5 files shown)


glrf (approved):

commodore/package/compile.py
```diff
@@ -63,7 +67,7 @@ def compile_package(
         cfg,
         targets,
         search_paths=[cfg.vendor_dir],
-        fetch_dependencies=True,
+        fetch_dependencies=cfg.fetch_dependencies,
```
corvus-ch, line 70: This can be dropped. The value can be access from the config object passed to the called context.

corvus-ch (changes requested):
Please address my remark of an obsolete parameter.

simu:
> Please address my remark of an obsolete parameter.

Clearly out-of-scope for this PR, as this would have to be changed in a bunch of places.

corvus-ch (approved):

commodore/package/compile.py
simu: Will be done in #508

PR #511, opened 2022-05-20
Change `package compile` target class to be provided as a path

simu:
Some packages may require that some configuration is provided in an additional class file, for example cluster facts may be used to include additional classes by the package.

Generally, users should define test cases in `tests/` in the package directory and have each test case class include the desired target class of the package in the test class's `classes` array. Note that relative includes are always interpreted relative to the class containing the include. Reclass supports the prefix `..` to include a class from the parent directory.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/package/compile.py
+++ b/commodore/package/compile.py
@@ -24,7 +24,6 @@ def compile_package(
 ):
     # Clean working tree before compiling package, some of our symlinking logic expects
     # that `inventory/` is cleaned before symlinks are created.
-    # TODO(sg): Make this conditional on not running local mode for package compilation
     if not cfg.local:
         clean_working_tree(cfg)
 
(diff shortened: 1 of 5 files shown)


commodore/package/compile.py
```diff
@@ -36,13 +35,13 @@ def compile_package(
...
-    # Verify root class
-    root_class_path = root_class.replace(".", "/")
-    root_cls = pkg_path / f"{root_class_path}.yml"
-    if not root_cls.is_file():
-        raise click.ClickException(f"Root class '{root_cls.stem}' doesn't exist.")
+    root_class_name = root_class.replace(".yml", "").replace("/", ".")
+    # Convert root class file to class name
+    root_class_path = pkg_path / root_class
+    if not root_class_path.is_file():
+        raise click.ClickException(f"Root class '{root_class_name}' doesn't exist.")
```
glrf, line 42: The naming is a bit inconsistent here. You call it test class in the documentation but root class in the error.

glrf (approved):
LGTM, just a tiny nitpick

commodore/package/compile.py
simu: I missed renaming that one... fixed now.

PR #515, opened 2022-05-25
Explicitly specify `oauthlib` as a dependency

simu:
We use `oauthlib` in `commodore/login.py` but didn't explicitly specify the dependency until now.

## Checklist


glrf (approved):
LGTM

I did not realize we imported that through Kapitan..

PR #544, opened 2022-07-07
component-template: Add `lint_kubent` make targets to check for deprecated API versions

simu:
The `lint_kubent` target runs `kubent` on a single golden test instance, defaulting to instance `defaults`. The commit also adds a target `lint_kubent_all` which will run `kubent` for all targets separately.

We don't run `kubent` over all golden tests simultaneously because the tool doesn't support linting multiple Kubernetes objects with the same name and namespace in a single run.

The commit ensures that the `kubent` lint targets are only added to components which are created with golden tests enabled.

See also https://github.com/projectsyn/modulesync-control/pull/82

## Checklist

- [x] Keep pull requests small so they can be easily reviewed.- [ ] Update the documentation.

### DIFF
--- a/commodore/component-template/{{ cookiecutter.slug }}/Makefile
+++ b/commodore/component-template/{{ cookiecutter.slug }}/Makefile
@@ -15,7 +15,7 @@ help: ## Show this help
 all: lint
 
 .PHONY: lint
-lint: lint_jsonnet lint_yaml lint_adoc ## All-in-one linting
+lint: lint_jsonnet lint_yaml lint_adoc{%- if cookiecutter.add_golden == "y" %} lint_kubent{% endif %} ## All-in-one linting
 
 .PHONY: lint_jsonnet
 lint_jsonnet: $(JSONNET_FILES) ## Lint jsonnet files
@@ -28,6 +28,12 @@ lint_yaml: ## Lint yaml files
 .PHONY: lint_adoc
 lint_adoc: ## Lint documentation
 	$(VALE_CMD) $(VALE_ARGS)
+
+{%- if cookiecutter.add_golden == "y" %}
+.PHONY: lint_kubent
+lint_kubent: ## Lint deprecated Kubernetes API versions
+	$(KUBENT_DOCKER) $(KUBENT_ARGS) -f $(KUBENT_FILES)
+{%- endif %}
 
 .PHONY: format
 format: format_jsonnet ## All-in-one formatting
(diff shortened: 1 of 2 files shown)


commodore/component-template/{{ cookiecutter.slug }}/Makefile
```diff
@@ -29,6 +29,12 @@ lint_yaml: ## Lint yaml files
 lint_adoc: ## Lint documentation
 	$(VALE_CMD) $(VALE_ARGS)
 
+{%- if cookiecutter.add_golden == "y" %}
+.PHONY: lint_kubent
+lint_kubent: ## Lint deprecated Kubernetes API versions
```
corvus-ch, line 34:
```suggestion
lint_kubent: ## Check for deprecated Kubernetes API versions
```

commodore/component-template/{{ cookiecutter.slug }}/Makefile.vars.mk
```diff
@@ -35,6 +35,16 @@ COMMODORE_CMD  ?= $(DOCKER_CMD) $(DOCKER_ARGS) $(root_volume) docker.io/projects
 COMPILE_CMD    ?= $(COMMODORE_CMD) component compile . $(commodore_args)
 JB_CMD         ?= $(DOCKER_CMD) $(DOCKER_ARGS) --entrypoint /usr/local/bin/jb docker.io/projectsyn/commodore:latest install
 
+{%- if cookiecutter.add_golden == "y" %}
+GOLDEN_FILES    ?= $(shell find tests/golden/$(instance) -type f)
+
+KUBENT_FILES    ?= $(shell echo "$(GOLDEN_FILES)" | sed 's/ /,/g')
```
corvus-ch, line 41: If I recall correctly, the length of a shell command is limited. Depending on the amount of golden files, that limit can be reached, can't it?

corvus-ch (approved):

commodore/component-template/{{ cookiecutter.slug }}/Makefile.vars.mk
simu: The length might be reached in some extreme cases, but it's imho unlikely, since modern Linux supports up to 128KiB **per argument** and up to 2MiB total for a command invocation.

We could just pass the files to kubent with `-f $(find tests/golden/$(instance) -type f -printf '%p,' | sed 's/,$//')` instead of precomputing the list in a make variable. I just reused the same pattern which we use for other make targets.

corvus-ch: Na, I am fine with having it in the variable.

Issue #518, opened 2022-05-25
Add support for dependencies in a subdirectory of a monorepo


simu:
## Context

Some software is developed in a monorepo which contains both the source code and any required configuration templates (e.g. Helm charts) required for deploying the software.

While it's technically already possible to use a config package in a subdirectory of a repository, it's not very ergonomic as any class includes of such a package would have format `<package-name>.path.in.repo.<class-name>`. For such software, it would be nice if Commodore's config package support would allow users to specify a path within the Git repository where the actual config package is stored. This would allow users to include classes of such a package using the usual `<package-name>.<class-name>` format.

Implementation-wise, this could be supported by introducing an additional key `path` for the package specifications in `parameters.packages`.

## Acceptance criteria

* Commodore supports an additional optional key `path` for dependency specifications. If the key is not present, the whole repo is used as the dependency, otherwise the subdirectory in the repo indicated by `path` is used as the dependency.
* Both components and packages in a repository subdirectory are supported

## Alternatives

Keep as is, let users deal with the inconvenient class names of such packages

## Implementation ideas

* Fetch repos into `packages` in the working directory and symlink actual package(s) to `inventory/classes`
* Fetch repo only once if multiple packages are stored in different subdirectories

simu:
Keep in mind that a repo might have multiple packages in different directories for implementation (i.e. deduplicate repo clones/fetches)

simu:
We'll have to somehow handle the case where multiple dependencies in the same monorepo are included with different versions. Either we'll have to clone the repo separately for each dependency, or we'll have to figure out how to make use of `git worktree`.

ccremer:
For a first iteration I'd be happy if the repo is shallow cloned for each dependency.
I don't know how it'll work with tags though.

Usage example how a monorepo release and tag scheme could look like:

* `vx.y.z` for normal code/image tags
* `<chart-name>-x.y.z` for chart version tags (which might be different than code tags, and there could be multiple charts)
* `component-x.y.z` for component version tags

The tags could be built up on eachother, e.g. `component-1.0.2` refers to a tag `chart-2.3.5`, which refers to docker image `v1.2.1`.

(A shallow clone of `component-x.y.z` would contain the full component with all the chart and image versions pinned)

However, another example of a monorepo would be consolidation of similar components and/or config packages for maintenance reasons. For example (tags):
- component-prometheus (`component-v2.4.5`)
[...]

PR #546, opened 2022-07-07
Clone config packages to `dependencies/pkg.{package-name}`

simu:
Instead of cloning packages directly into `inventory/classes` we now clone them to `dependencies/pkg.{package-name}` and symlink them into `inventory/classes`.

This is a first step towards supporting dependencies in repository subdirectories, and unifies the location for cloned component and package repositories in the Commodore working directory.

Part of #518 

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/package/compile.py
+++ b/commodore/package/compile.py
@@ -20,15 +20,9 @@ from commodore.inventory import Inventory
 from commodore.postprocess import postprocess_components
 
 
-# pylint: disable=too-many-arguments disable=too-many-locals
-def compile_package(
-    cfg: Config,
-    pkg_path_: str,
-    root_class: str,
-    value_files_: Iterable[str],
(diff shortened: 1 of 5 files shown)


commodore/package/compile.py
```diff
@@ -20,15 +20,9 @@
...
-# pylint: disable=too-many-arguments disable=too-many-locals
-def compile_package(
-    cfg: Config,
-    pkg_path_: str,
-    root_class: str,
-    value_files_: Iterable[str],
-    tmp_dir: Optional[str] = "",
-    keep_dir: bool = False,
-):
+def set_work_dir(
```
corvus-ch, line 23:
```suggestion
def create_and_set_work_dir(
```

Would be a bit more expressive.

corvus-ch (approved):

PR #547, opened 2022-07-08
Add support for dependencies in repository subpaths

simu:
This PR adds support for storing dependencies (components and packages) in a subpath of a repository. This is a prerequisite for supporting repositories which contain multiple dependencies.

Currently, Commodore will clone a repository multiple times if multiple dependencies are stored in the same repository. We'll introduce repository deduplication in a follow-up PR (#549).

Part of #518 

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/dependency_mgmt/__init__.py
+++ b/commodore/dependency_mgmt/__init__.py
@@ -37,16 +37,19 @@ def create_component_symlinks(cfg, component: Component):
         )
 
 
-def create_package_symlink(cfg, pname: str, package: Package):
+def create_package_symlink(cfg, pname: str, package: Package, path: str):
     """
     Create package symlink in the inventory.
 
-    Packages are downloaded to `dependencies/pkg.{package-name}` and symlinked to
+    Packages are downloaded to `dependencies/pkg.{package-name}/{path}` and symlinked to
     `inventory/classes/{package-name}`.
     """
     if not package.target_dir:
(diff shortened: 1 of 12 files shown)


commodore/dependency_mgmt/__init__.py
```diff
@@ -37,16 +37,19 @@ def create_component_symlinks(cfg, component: Component):
...
+    Packages are downloaded to `dependencies/pkg.{package-name}/{path}` and symlinked to
     `inventory/classes/{package-name}`.
     """
     if not package.target_dir:
         raise ValueError("Can't symlink package which doesn't have a working directory")
-    relsymlink(package.target_dir, cfg.inventory.classes_dir, dest_name=pname)
+    package_path = package.target_dir / path
+    if not package_path.is_dir():
+        raise ValueError(f"Can't symlink package subpath {path}, not a directory")
+    relsymlink(package_path, cfg.inventory.classes_dir, dest_name=pname)
```
glrf, line 52: Is there a reason that that the sub_path was integrated in the component class but handled separately for packages?

glrf (approved):
LGTM

commodore/dependency_mgmt/__init__.py
simu: The `Package` class doesn't provide any properties or methods which require the sub_path, but it probably makes sense to add it to the class anyway.

PR #556, opened 2022-07-15
Create new components and packages as Git worktree checkouts

simu:
Only create worktrees when creating components or packages in a Commodore working tree (i.e. as subdirectories of `dependencies/`).
    
We also change the generic dependency template rendering to always render the template in a temp directory and moving the result into the final location, because we can't make an existing directory a Git worktree and cookiecutter doesn't support rendering a template into an existing directory, so we need to create the worktree first, then copy the templated dependency into it.

Follow-up to #551 

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/component/template.py
+++ b/commodore/component/template.py
@@ -83,6 +83,17 @@ class ComponentTemplater(Templater):
                     abort=True,
                 )
             rmtree(component.target_directory)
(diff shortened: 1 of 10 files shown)


glrf (approved):
LGTM

PR #559, opened 2022-07-19
Support different versions per instance of a component

### LINKED ISSUE #563: Allow users to specify different versions for different instances of the same component
## Context

We have some setups where we run multiple instances of the same component on a single cluster where it would be desirable to be able to update the component version for each instance separately. Currently, Commodore will use the same version for each component instance, but with the new dependency checkout structure based on worktrees we're able to checkout different versions of a component for different component instances.

## Current state

In #559, we provide an implementation which allows suitably written components to be rendered with different versions for different instances. However, there's no logic yet which verifies whether a component will work correctly when checked out in different versions for different instances. See the PR description for a brief overview over the current state of the implementation.

## Missing features

* We can't support different versions of Jsonnet dependencies for different instances of the same component with the current Jsonnet dependency management architecture. In the draft PR, Commodore makes the Jsonnet dependencies of the "main" non-instantiated component checkout available for all instances. We could possibly achieve support for separate Jsonnet dependency versions per instance, but it would require reworking how we interface with Kapitan's `compile()` function.
* Components **must** use `${_base_directory}` in their compile configuration (`parameters.kapitan.dependencies` and `parameters.kapitan.compile`) in order to ensure Helm & HTTPS dependencies are fetched individually for each instance and that the instance's Jsonnet files are picked up by `kapitan compile`. The draft PR doesn't perform any validation when rendering component instances of the same component with different versions. There's multiple options to handle this: 1) we simply require that compnents advertise instance-versioning support 2) we lint the component classes to check that there's no entries which don't use `${_base_directory}` (TBD whether that's feasible for all cases).
* The architecture of specifying versions for instances theoretically allows changing the repository URL and sub-path in the repository per instance. However, the implementation in the draft PR doesn't support this yet, as it's unclear how this needs to be modelled in the internal component/alias architecture in Commodore.

## Links

* #559

simu:
This PR changes Commodore to

* Create separate worktrees for each component alias
* Create class symlinks for each component alias
* Create each alias target with only the defaults and component class symlinked from the alias worktree
* Read instance versions from `parameters.components.<instance-name>`

Note that per-instance versions only work correctly only for components which use `${_base_directory}` in their config when specifying Jsonnet files or Helm chart/YAML locations in `kapitan.compile` and `kapitan.dependencies`. Note, that components should use `${_base_directory}` anyway, and new components created from the template use `${_base_directory}` out of the box.

TODO:
- [x] Implement support for overriding the URL or path for individual component instances
- [x] Test cases which cover new features

Resolves #563 

## Checklist

- [x] Update tests.


commodore/dependency_mgmt/__init__.py
```diff
@@ -211,9 +216,8 @@ def register_components(cfg: Config):
 
         c = registered_components[cn]
         aspec = cspecs[alias]
-        if aspec.url != c.repo_url or aspec.path != c.sub_path:
-            raise NotImplementedError("Changing alias sub path / URL NYI")
-        c.register_alias(alias, aspec.version)
+        if aspec.url == c.repo_url and aspec.path == c.sub_path:
+            c.register_alias(alias, aspec.version)
```
simu, line 220: This function is used when running Commodore in "local mode" (`--local`). From my quick local test this will now fail with a less comprehensible error than the original draft's `NotImplementedError` since we still don't handle changed URL/path here.

We probably want something like

```suggestion
        if aspec.url != c.repo_url:
            adep = cfg.register_dependency_repo(aspec.url)
            adep.register_component(alias, component_dir(cfg.work_dir, alias))
        c.register_alias(alias, aspec.version, aspec.path)
```

commodore/component/__init__.py
```diff
@@ -126,21 +134,44 @@ def sub_path(self) -> str:
...
+
+    def alias_directory(self, alias: str) -> P:
+        if not self._dependency:
+            return self._dir / self._sub_path
+        apath = self._dependency.get_component(alias)
+        if not apath:
+            raise ValueError(f"unknown alias {alias} for component {self.name}")
+        if alias in self._aliases:
+            return apath / self._aliases[alias][1]
+        return apath / self._sub_path
```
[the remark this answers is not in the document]
HappyTetrahedron: Given that removing this line fails a bunch of tests I'd say it's needed.
It's possible that the tests simply set things up wrong and fail the precondition, but I also don't particularly want to fix 98 tests.

docs/modules/ROOT/pages/reference/architecture.adoc
```diff
@@ -299,6 +303,52 @@ Commodore currently doesn't provide support for component authors to specify lib
 It's the responsibility of component authors to agree on an interface and to ensure that their implementations adhere to the interface.
 ====
 
+=== Component instance versions
```
simu, line 306: I'd move this section above `=== Component dependencies` so it comes directly after `=== Component instances`

Note that I've renamed it to better match the other section titles

docs/modules/ROOT/pages/reference/architecture.adoc
```diff
@@ -60,6 +60,10 @@ For each dependency which is stored in this repository, Commodore ensures that t
 Regardless of the value of key `path`, Commodore creates a checkout of the  complete repository in `dependencies/<dependency-name>`.
 However, Commodore will create a symlink to the specified path when making the dependency available in the hierarchy.
 
+For components which are instantiated multiple times, Commodore ensures that an additional Git worktree in `dependencies/<instance-name>` exists for each component instance and is checked out to the instance's desired component version.
```
simu, line 63: I quickly added this section since we now create a Git worktree for each component instance regardless of whether a component supports per-instance versions.

simu (commented):
LGTM now (once my final comments are addressed).

docs/modules/ROOT/pages/reference/architecture.adoc
```diff
@@ -299,6 +299,52 @@ Commodore currently doesn't provide support for component authors to specify lib
 It's the responsibility of component authors to agree on an interface and to ensure that their implementations adhere to the interface.
 ====
 
+=== Using different component versions across instances
+
+With https://syn.tools/syn/SDDs/0033-commodore-component-instance-versioning.html[SDD0033], we've introduced support for instantiating components with different versions, meaning that different instances of the same component may use a different dependency or different versions of the same dependency.
+Component authors must explicitly declare that their component supports multi-version instantiation.
+Components advertise that they support multi-version instantiation by setting the field `multi_version` in `parameters.<component_name>._metadata` to `true`.
```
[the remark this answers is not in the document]
HappyTetrahedron: Looks like my brain also prefers multi_version - I was so certain that was correct, I'd decided I don't need to bother checking.
I changed it in the code.

simu:
I've updated the SDD to match the final implementation: https://github.com/projectsyn/documentation/pull/184

HappyTetrahedron (approved):

PR #565, opened 2022-07-21
Add flag for providing fallback dynamic facts for `catalog compile`

simu:
We support providing dynamic facts as `--dynamic-fact key=value` in command `catalog compile`. The key can specify a nested key as `path.to.key`, and the value can optionally be parsed as JSON if it's prefixed with a literal `json:`. Facts specified later on the command line take precedence over previously specified facts. Nested facts will not be considered if one of their key segments (except the leaf) already holds a non-dictionary value.

Dynamic facts provided on the command line are only used for catalog compilation when the Lieutenant API response for the cluster which is getting compiled doesn't contain field `dynamicFacts`.

This enables us to cleanly support compilation of a cluster catalog before Steward is deployed on the cluster.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/cluster.py
+++ b/commodore/cluster.py
@@ -20,10 +20,18 @@ from .inventory import Inventory
 class Cluster:
     _cluster_response: dict
     _tenant_response: dict
+    _fallback_dynamic_facts: dict[str, Any]
 
-    def __init__(self, cluster_response: dict, tenant_response: dict):
+    def __init__(
+        self,
+        cluster_response: dict,
+        tenant_response: dict,
+        fallback_dynamic_facts: dict[str, Any] = {},
+    ):
         self._cluster = cluster_response
         self._tenant = tenant_response
+        self._fallback_dynamic_facts = fallback_dynamic_facts
+
         if (
             "tenant" not in self._cluster
             or self._cluster["tenant"] != self._tenant["id"]
@@ -95,7 +103,7 @@ class Cluster:
 
     @property
     def dynamic_facts(self) -> dict[str, Any]:
-        return self._cluster.get("dynamicFacts", {})
+        return self._cluster.get("dynamicFacts", self._fallback_dynamic_facts)
 
 
 def load_cluster_from_api(cfg: Config, cluster_id: str) -> Cluster:
@@ -107,7 +115,7 @@ def load_cluster_from_api(cfg: Config, cluster_id: str) -> Cluster:
     tenant_response = lieutenant_query(
         cfg.api_url, cfg.api_token, "tenants", cluster_response["tenant"]
     )
-    return Cluster(cluster_response, tenant_response)
+    return Cluster(cluster_response, tenant_response, cfg.dynamic_facts)
 
 
 def read_cluster_and_tenant(inv: Inventory) -> tuple[str, str]:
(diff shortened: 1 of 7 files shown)


commodore/config.py
```diff
@@ -282,3 +294,119 @@ def _component_is_aliasable(cluster_parameters: dict, component_name: str):
...
+            # nested key was malformed.
+            continue
+
+        if key in target_fact:
+            click.secho(
+                " > Overwriting dynamic fact "
+                + f"{raw_key}={target_fact[key]} with '{value}'",
+                fg="yellow",
+            )
+        target_fact[key] = value
```
glrf, line 410: Nitpick: I found the `parse_sub_key` function fairly confusing at first as I wasn't expecting it to return a sub dict and was confused how the actual dict was updated. I would probably done something like `set_fact_value(facts, raw_key, value)` and do both key parsing an setting the value in the helper function, but that's just personal preference so :shrug:

commodore/cluster.py
```diff
@@ -95,7 +103,7 @@ def facts(self) -> dict[str, str]:
 
     @property
     def dynamic_facts(self) -> dict[str, Any]:
-        return self._cluster.get("dynamicFacts", {})
+        return self._cluster.get("dynamicFacts", self._fallback_dynamic_facts)
```
glrf, line 106: If I read this correctly we use the fallback dynamic facts only if the API returns no `dynamicFacts` field, I would probably suggest to also use the fallback facts if the returned fact dict is empty.

The current works with the current API implementation, but as far as I understand the openAPI definition https://github.com/projectsyn/lieutenant-api/blob/master/openapi.yaml returning an empty fact dict would also be valid and sounds like something that could be changed without realizing that it would be a breaking change.

This will most likely never happen, so feel free to ignore this suggestions.

glrf (approved):
LGTM :rocket: 

I'm not 100% sure that there is not some edge case bug hiding somewhere in the parsing , but I think this is good enough. (Might be an opportunity to play with fuzz testing though)

commodore/config.py
simu: Probably makes more sense since the diagnostic messages in `parse_sub_key()` don't really make sense if the function was used somewhere else.

commodore/cluster.py
simu: I originally explicitly only used the manually provided dynamic facts when there's no field `dynamicFacts` in the API response, to minimize the potential for users to be able to push non-reproducible commits to the cluster catalog by providing incorrect/invalid dynamic facts on the command line for clusters which are bootstrapped but don't report any facts (note that this scenario is currently purely hypothetical as well).

Thinking about it again, I don't really have a strong opinion on whether to allow users to provide arbitrary facts even if the API response for the cluster contains an empty `dynamicFacts` dict.

glrf: That's also a valid point. I'm not sure anymore which approach is better.

Maybe we could print a warning if Lieutenant returns dynamic facts (event empty ones) and the user tries to provide their own? That way we wouldn't silently fail if the API starts to return `{}`, making debugging a bit easier.

glrf (approved):

simu:
> I'm not 100% sure that there is not some edge case bug hiding somewhere in the parsing , but I think this is good enough. (Might be an opportunity to play with fuzz testing though)

Side-note: I did try out some fuzzing tools for Python, but didn't find anything that can be easily integrated in our existing pytest setup, so I just ran the fuzzer locally. I wasn't able to get it to crash the parsing so far, so I'm reasonably happy with the robustness of the code.

PR #567, opened 2022-07-27
Add CLI commands to extract package versions and show rendered inventory

glrf:
This PR introduces tow new command for the command group `inventory`.

The command `inventory packages` is the equivalent to `inventory components`. It calls reclass to render the inventory of a global defaults or tenant repo and returns all packages (URLs and versions) defined in that repo.

The command `inventory show` also renders the inventory, but returns all parameters, not just package or component infos.

Both commands can take an extra values file and can return the output in json or yaml

## Checklist

- [x] Update tests.

### DIFF
--- a/docs/modules/ROOT/pages/reference/commands.adoc
+++ b/docs/modules/ROOT/pages/reference/commands.adoc
@@ -81,16 +81,18 @@ configuration repositories.
 
 The option `--alias` (or `-a`) can be used to compile an instance-aware component with a specific instance.
 
-== Inventory Components
+== Inventory Show
 
-  commodore inventory components GLOBAL_CONFIG [TENANT_CONFIG]
+  commodore inventory show|components|packages GLOBAL_CONFIG [TENANT_CONFIG]
 
-This command lists all the components defined in `parameters.components` in the Commodore hierarchy in directory `GLOBAL_CONFIG`.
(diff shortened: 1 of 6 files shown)


docs/modules/ROOT/pages/reference/commands.adoc
```diff
@@ -81,16 +81,18 @@ configuration repositories.
...
 
-  commodore inventory components GLOBAL_CONFIG [TENANT_CONFIG]
+  commodore inventory show|components|packages GLOBAL_CONFIG [TENANT_CONFIG]
 
-This command lists all the components defined in `parameters.components` in the Commodore hierarchy in directory `GLOBAL_CONFIG`.
+The command `commodore inventory show` prints the `parameters` in the Commodore hierarchy in directory `GLOBAL_CONFIG`.
 If provided, the command also takes into account the tenant repo in directory `TENANT_CONFIG`.
 
-NOTE: The command doesn't currently support cloning either the global or tenant repository from a Git repo URL.
+The commands `commodore inventory components` and `commodore inventory packages` also render the inventory, but only print a list of all components or packages.
```
simu, line 91: This feels more natural to me, but I'm not sure if it's actually more correct:

```suggestion
The commands `commodore inventory components` and `commodore inventory packages` also render the inventory, but only print the list of all components and packages respectively.
```

simu (approved):

PR #582, opened 2022-08-11
Deprecate parameter `--repo-directory` for component compile

simu:
We don't need users to provide the location of the repo directory in `component compile` since we strictly speaking don't need the component to even be in a Git repository for standalone compilation. We  use GitPython's `search_parent_directories=True` to try to find the repo in which the provided target dir is located, but fall back on just using the indicated target directory as the component directory for standalone compilation.

If we find a Git repository holding the component, we compute the component subpath based on the relative location of the target directory to the repo working tree directory. In this case, we use the working tree directory of the repo as the component "target directory" so that standalone compilation of a component stored in a repo subdirectory is as close as possible to compiling the component in a cluster catalog.

We don't remove the `--repo-directory` parameter, but print a deprecation warning when users provide it, and ignore the provided value in the implementation.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/cli.py
+++ b/commodore/cli.py
@@ -415,9 +415,7 @@ def component_delete(config: Config, slug, force, verbose):
     default="",
     show_default=False,
     type=click.Path(file_okay=False, dir_okay=True),
-    help="Path to the root of the Git repo containing the component. "
-    + "The command assumes that the component is in the repo root, "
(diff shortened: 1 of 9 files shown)


bastjan (approved):
I don't know the code well enough to thoroughly review it but the tests make sense to me.

PR #610, opened 2022-08-25
Add fallback logic for component template `add_pp` and `add_lib` cookiecutter args

simu:
Check the component implementation to determine whether the component has postprocessing filters and a component library if flags `add_pp` and `add_lib` aren't present in `.cruft.json`. This allows us to migrate to Cruft from modulesync easier as we don't have enough info in modulesync to set these flags.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/component/template.py
+++ b/commodore/component/template.py
@@ -5,6 +5,7 @@ from shutil import rmtree
 
 import click
 import git
+import yaml
 
 from commodore.config import Config
 from commodore.component import Component, component_dir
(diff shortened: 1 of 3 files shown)


commodore/component/template.py
```diff
@@ -21,12 +22,47 @@ class ComponentTemplater(Templater):
     def from_existing(cls, config: Config, path: Path):
         return cls._base_from_existing(config, path, "component")
 
+    @property
+    def _has_lib(self) -> bool:
+        """Determine whether component has postprocessing filters by checking whether
```
glrf, line 27:
```suggestion
        """Determine whether component has a component library by checking whether
```

commodore/dependency_templater.py
```diff
@@ -85,7 +85,26 @@ def _base_from_existing(cls, config: Config, path: Path, deptype: str):
         t._target_dir = path
         t.output_dir = path.absolute().parent
 
-        t._initialize_from_cookiecutter_args(cookiecutter_args)
+        # We pass the cookiecutter args dict to `_initialize_from_cookiecutter_args()`.
+        # Because Python dicts are passed by reference, the function can simply add
+        # missing args into the dict and return `True` to cause us to write back and
+        # commit the updated `.cruft.json` data.
+        update_cruft_json = t._initialize_from_cookiecutter_args(cookiecutter_args)
```
glrf, line 92: This is a bit too much magic for my taste, but I don't have a better solution. But I'd propose that at least add a doc string to `_initialize_from_cookiecutter_args` so that when someone looks at the method (or tried to extend the templater) they know why it returns a boolean and what it means without having to read this method.

glrf (approved):
LGTM overall

commodore/dependency_templater.py
simu: I'm not particularly happy with the amount of magic here either, but less magic would mean that we need to have a bunch of additional repetitive code. I'll add a docstring to explain the magic which happens.

PR #614, opened 2022-08-29
Add fallback to extract copyright info from modulesync config

simu:
If we don't have the copyright info for a component in the Cruft cookiecutter args, fall back to reading it from the component's
modulesync config (`.sync.yml`) or use the default values used by modulesync, if the component's `.sync.yml` doesn't have the information either.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/component/template.py
+++ b/commodore/component/template.py
@@ -57,6 +57,25 @@ class ComponentTemplater(Templater):
             cookiecutter_args["add_pp"] = "y" if self._has_pp else "n"
             update_cruft_json = True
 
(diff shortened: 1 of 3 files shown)


simu:
Side-note: this is necessary, because we don't have the required information available in modulesync when rendering the initial `.cruft.json`, because that data was scoped to only the LICENSE file in the modulesync template, cf. https://github.com/projectsyn/modulesync-control/pull/85#discussion_r956999316

glrf (approved):

PR #633, opened 2022-09-26
Modify `GitRepo.stage_all()` to respect `.gitignore`

### LINKED ISSUE #626: `commodore component update` commits files in gitignore

  1. Have files that are in `.gitignore`. e.g. by running `make golden-diff`
  1. Run `commodore component update .`
  1. Observe that directories such as `vendor` and `compiled` are commited

## Actual Behavior

The component update command (and I suspect also the package update command) does not look at the `.gitignore` .

## Expected Behavior

Files in `.gitignore` are not commited by `commodore component update`

simu:
We change the implementation of `stage_all()` to use the results of `git.Repo.untracked_files` and `git.index.diff(None)` to stage all changes instead of `Repo.index.add('*')`.

This change ensures that we respect the contents of `.gitignore` when staging changes with `stage_all()` as both `git.Repo.untracked_files` calls out to `git status` and therefore takes into account `.gitignore`, and `git.index.diff(None)` will never consider untracked files and thus won't ever cause untracked ignored files to be staged.

Fixes #626

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/gitrepo.py
+++ b/commodore/gitrepo.py
@@ -547,22 +547,17 @@ class GitRepo:
         to_add = self._repo.untracked_files
         # We don't want to remove anything by default.
         to_remove = []
-        # Stage deletions and add changes to `to_add`
+
+        # Determine changes to stage, separated into removals and other changes
         changes = self._repo.index.diff(None)
         if changes:
-            for c in changes.iter_change_type("D"):
-                # Track removed files for `index.remove()`
-                to_remove.append(c.b_path)
-
-            for c in changes.iter_change_type("M"):
-                # Track modified files for `index.add()`
-                to_add.append(c.a_path)
-            for c in changes.iter_change_type("T"):
-                # Track files with mode changes (e.g. regular file to symlink) for
-                # `index.add()`
-                to_add.append(c.a_path)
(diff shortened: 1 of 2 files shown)


glrf (approved):

commodore/gitrepo.py
```diff
@@ -539,6 +539,33 @@ def _check_conflicts(self):
...
+        the index."""
+        # We always want to stage untracked files. The implementation for
+        # `untracked_files` respects the repo's `.gitignore`.
+        to_add = self._repo.untracked_files
+        # We don't want to remove anything by default.
+        to_remove = []
+        # Stage deletions and add changes to `to_add`
+        changes = self._repo.index.diff(None)
+        if changes:
+            for c in changes.iter_change_type("D"):
```
corvus-ch, line 553: Does the library provide constants for the change type? If so, then I would use them instead.

corvus-ch (approved):

commodore/gitrepo.py
simu: There's no constants for the change types, otherwise the code would probably use them already.

corvus-ch: I thought so, but you never know.

PR #635, opened 2022-09-27
 Change `sync_dependencies()` to omit PRs which only update the template commit id

### LINKED ISSUE #632: Commodore's template sync creates PRs even if there's no actual changes
Commodore's template sync creates essentially empty PRs when ran after a change to the template which only affects a few dependencies, see e.g. https://github.com/projectsyn/component-cert-manager/pull/105. This is not ideal, and we should ensure that we only create PRs for dependencies which have actual changes (other than the template repo commit ID).

  1. Make a template change which only applies to a subset of dependencies (e.g. https://github.com/projectsyn/commodore-component-template/pull/18)
  1. Observe that the template sync logic creates a PR for components which has no changes (e.g. https://github.com/projectsyn/component-cert-manager/pull/105)

## Actual Behavior
The template sync interprets the changed template commit ID as a change in the dependency's boilerplate. While technically true, it's not worth it to actually create a PR which only bumps the template commit ID.

## Expected Behavior

The template sync only creates PRs for dependencies which have non-trivial changes.

simu:
This PR adds an optional argument for `Templater.update()` which configures a custom diff function which suppresses `.cruft.json` diffs which only contain a changed template commit id.

This allows us to change `sync_dependencies()` to call `update()` with `ignore_template_commit=True` to ensure the sync doesn't create PRs for updates which only modify the template commit id for template updates which make changes in a part of the template which only applies to a few dependencies.

Fixes #632 

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/dependency_templater.py
+++ b/commodore/dependency_templater.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import datetime
+import difflib
 import json
 import re
 import tempfile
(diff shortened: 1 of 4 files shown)


commodore/dependency_templater.py
```diff
@@ -15,12 +16,48 @@
...
+                tofile=tofile,
+                n=0,
+            )
+        )[3:]
+        # If the context-less diff has exactly 2 lines and both of them start with
+        # '[-+] "commit":', we omit the diff
+        if (
+            len(minimal_diff) == 2
+            and minimal_diff[0].startswith('-  "commit":')
+            and minimal_diff[1].startswith('+  "commit":')
```
glrf, line 54: I suspect it's always guaranteed that the `-` line is before the `+` line?

glrf (approved):

commodore/dependency_templater.py
simu: Yes, that's my understanding of the unified diff format returned by `difflib.unified_diff()`

PR #646, opened 2022-10-07
Improve support for `kustomize` in Commodore

### LINKED ISSUE #504: Add support for using `kustomize` in components
## Context

We've written a bunch of components where we manually download upstream manifests which are intended to be rendered with kustomize, e.g. https://github.com/projectsyn/component-kyverno/. By having to download and render all manifests individually instead of being able to leverage kustomize, errors can happen, e.g. https://github.com/projectsyn/component-kyverno/pull/46

There's already a draft PR to add `kustomize` to the Commodore container image: #430

Additionally, we'll want at least basic documentation showing how to call kustomize from a component. We may be able to use Kapitan's [external input type](https://kapitan.dev/compile/#external-alphaexperimental) to run kustomize. This will require the standalone `kustomize` binary to be present in the path when Commodore is running.

## Alternatives

* Continue downloading kustomize manifests and patching them in Jsonnet

simu:
Provide the absolute path to the script in reclass variable `${_kustomize_wrapper}`, to allow users to make use of Kustomize without having to worry about the binary location being different on different systems.

The PR also provides a new function in `commodore.libjsonnet`, which generates an object which is suitable as a Jsonnet output to render a `kustomization.yaml` which can be processed by `kustomize build`.

See https://github.com/projectsyn/documentation/pull/169 for extended documentation.

Closes #504 

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/component/compile.py
+++ b/commodore/component/compile.py
@@ -12,6 +12,7 @@ import click
 import git
 from kapitan.resources import inventory_reclass
 
+from commodore import __kustomize_wrapper__
 from commodore.config import Config
 from commodore.component import Component
 from commodore.dependency_mgmt.component_library import (
@@ -250,6 +251,7 @@ def _prepare_kapitan_inventory(
             "parameters": {
                 "_instance": instance_name,
                 "_base_directory": str(component.target_directory),
+                "_kustomize_wrapper": str(__kustomize_wrapper__),
             },
         },
         inv.target_file(instance_name),
--- a/commodore/__init__.py
+++ b/commodore/__init__.py
@@ -10,3 +10,6 @@ __version__ = version("syn-commodore")
 
 # provide Commodore installation dir as variable that can be imported
 __install_dir__ = P(__file__).parent
+
+# Location of Kustomize wrapper script
+__kustomize_wrapper__ = __install_dir__ / "scripts" / "run-kustomize"
--- a/commodore/cluster.py
+++ b/commodore/cluster.py
@@ -6,6 +6,7 @@ from typing import Any, Optional, Union
 
 import click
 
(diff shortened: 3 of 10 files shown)


commodore/component/compile.py
```diff
@@ -250,6 +251,7 @@ def _prepare_kapitan_inventory(
             "parameters": {
                 "_instance": instance_name,
                 "_base_directory": str(component.target_directory),
+                "_kustomize_wrapper": str(__kustomize_wrapper__),
```
bastjan, line 254: Would it be complicated to also add `${_compile_directory}` or `${_output_directory}` or similar?

bastjan (approved):

commodore/component/compile.py
simu: It's possible but will be fragile, since the compile target directory is computed by Kapitan based on the target we're defining here (and in `render_target()` for `catalog compile).

See https://github.com/kapicorp/kapitan/blob/257f53962d375be12da74d9c966e6e423aa68a90/kapitan/targets.py#L420-L468 and https://github.com/kapicorp/kapitan/blob/257f53962d375be12da74d9c966e6e423aa68a90/kapitan/inputs/base.py#L83-L87

simu: Actually Kapitan creates a temp directory (cf. https://github.com/kapicorp/kapitan/blob/257f53962d375be12da74d9c966e6e423aa68a90/kapitan/targets.py#L50-L54) which is used as the output directory during compilation, and copies the resulting tree into the final output directory after compilation is completed. We can't really infer the temp directory location ahead of time, so this isn't really feasible in a robust fashion, since the final output directory may not exist at the point when `kustomize` is called.

Issue #659, opened 2022-10-21
`commodore inventory lint` crashes on YAML documents which contain top-level `parameters` key which isn't a dict


simu:
Adding a YAML file which starts with a document separator, such as

```
parameters:
  - name: FOO
    value: bar
```

to an inventory causes Commodore to crash with

```
  File "/home/simon/.local/commodore-venv/lib/python3.8/site-packages/commodore/inventory/lint_dependency_specification.py", line 14, in lint_dependency_specification
    components = filecontents.get("parameters", {}).get(deptype.value, {})
AttributeError: 'list' object has no attribute 'get'
```

## Steps to Reproduce the Problem

  1. Add the snippet shown above as a YAML file in the inventory
  1. Run `commodore inventory lint`
  1.

## Actual Behavior

Commodore crashes

## Expected Behavior

Commodore prints a warning that it found a file in which {{parameters:}} isn't a dictionary

Optionally, we may want to instead add support for specifying files to ignore in the linter

PR #660, opened 2022-10-21
Avoid crash in `inventory lint` when top-level `parameters` is not an object

simu:
Until now, we didn't check whether a top-level key `parameters` is actually a dict, which caused a crash in the linter when a non-inventory file (e.g.  an OpenShift template) has such a top-level key.

This commit simply skips files which have a top-level key `parameters` which isn't a dict.

Part of #659

## Checklist

- [x] Keep pull requests small so they can be easily reviewed
- [x] Update tests.


simu:
I'm overriding review rules so the VSHN CI pipelines (which use the `:latest` tag) are working again for the tenant repo where we discovered the bug.

PR #665, opened 2022-10-25
Implement ignore patterns for `inventory lint`

simu:
Patterns can be provided either on the command line in `--ignore-patterns` or in a `.commodoreignore` file in the linting
target path. Patterns given on the command line apply for each target path, while `.commodoreignore` applies only for the target path in which the file is present.

Patterns are expanded recursively using Python's `glob` library.  You can use `*`, `?`, and character ranges expressed as `[]` with the usual semantics of shell globbing, and `**` to indicate an arbitrary amount of subdirectories. Patterns which start with `/` are treated as anchored in the target path, while other patterns are treated as matching any subpath in the target path.

Part of #659 

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/inventory/lint.py
+++ b/commodore/inventory/lint.py
@@ -26,28 +26,28 @@ class LintFunc(Protocol):
 class Linter:
     @abc.abstractmethod
     def __call__(
-        self, config: Config, path: Path, ignore_patterns: tuple[str] = ()
+        self, config: Config, path: Path, ignore_patterns: tuple[str, ...] = ()
     ) -> int:
         ...
 
 
 class ComponentSpecLinter(Linter):
     def __call__(
-        self, config: Config, path: Path, ignore_patterns: tuple[str] = ()
+        self, config: Config, path: Path, ignore_patterns: tuple[str, ...] = ()
     ) -> int:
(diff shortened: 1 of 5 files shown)


commodore/inventory/lint.py
```diff
@@ -92,14 +103,56 @@ def _lint_directory(cfg: Config, path: Path, lintfunc: LintFunc) -> int:
...
+) -> set[Path]:
+    """Use `glob.glob()` to render the paths to ignore from `ignore_patterns`.
+    All patterns are rooted at `base_dir`.
+
+    If pattern doesn't start with /, we will treat it as matching any prefix in
+    `base_dir`."""
+
+    # Read additional ignore patterns from `<base_dir>/.commodoreignore`, if the file
+    # exists
+    _ignore_patterns = set(p for p in ignore_patterns)
```
glrf, line 138: Maybe this is some python thing I don't understand but why the list comprehension?

glrf (approved):

commodore/inventory/lint.py
simu: This was a leftover from some previous version of the implementation. It's not required at all.

PR #672, opened 2022-10-27
Refactor `cli.py` into a package

simu:
Make all command flags which we use more than once reusable

## Checklist


glrf (approved):

simu:
Changes since review: I moved each set of sub-commands from `cli/__init__.py` to their own files in the `cli` package and link them to the `commodore` group with `add_command()` explicitly in `__init__.py`. This simplifies finding relevant CLI entrypoint functions when working on a particular part of Commodore.

This also required some changes for the CLI function tests, which I've also moved to new test files for the respective files in the `cli` package.

glrf (approved):
LGTM, it's a lot more readable that way IMO

PR #686, opened 2022-11-15
Fix mounting of GPG agent to Commodore docker container on Linux

mweibel:
without the change the gpg opts would be passed to `docker run` as a single argument instead of two separate arguments.

Not sure if there is a better option to do this or in which shell/version combination this should've worked :) 

## Checklist

- [ ] Update tests.

### DIFF
--- a/docs/modules/ROOT/pages/explanation/running-commodore.adoc
+++ b/docs/modules/ROOT/pages/explanation/running-commodore.adoc
@@ -115,7 +115,7 @@ On Linux it's possible to use SSH agent and mounting the agents socket into the 
 commodore() {
   local pubring="${HOME}/.gnupg/pubring.kbx"
   if command -v gpgconf &>/dev/null && test -f "${pubring}"; then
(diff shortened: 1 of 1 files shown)


simu:
> Not sure if there is a better option to do this or in which shell/version combination this should've worked :)

Counter question: on which shell(s) does your change work?

simu (approved):
New command works fine for Linux+bash

mweibel:
> > Not sure if there is a better option to do this or in which shell/version combination this should've worked :)
> 
> Counter question: on which shell(s) does your change work?
> 
> IIRC I verified that the old command worked with bash on Linux

tested on 
```GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)```

simu:
Integration, coverage and PyPI checks fail because they don't have access to the actions secrets for PRs from a fork. See #687 for a mild improvement. I've removed coverage and integration from the required checks in the branch protection config, to allow us to merge PRs from forks without having to jump through too many hoops.

PR #689, opened 2022-11-15
Fix list indentation if using post-process filters

### LINKED ISSUE #655: YAML generated by Commodore postprocess filters uses non-standard formatting
Commodore renders the YAML for postprocessing filter outputs with the default PyYAML renderer. Unfortunately, this renderer doesn't indent list entries the same way that's commonly accepted, and creates outputs like

```yaml
list:
- a
- b
``` 

instead of

```yaml
list:
  - a
  - b
```

  1. Compile a catalog or component which uses postprocessing filters
  1. Inspect the lists in the postprocessed output YAMLs

## Actual Behavior

YAML lists are rendered as follows after postprocessing:

```yaml
list:
- a
- b
``` 

## Expected Behavior

YAML lists are rendered as follows after postprocessing:

```yaml
list:
  - a
  - b
```

bastjan:
Fixes #655 

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/catalog.py
+++ b/commodore/catalog.py
@@ -14,7 +14,13 @@ import json
 
 from .component import Component
 from .gitrepo import GitRepo, GitCommandError
-from .helpers import ApiError, rm_tree_contents, lieutenant_query, sliding_window
+from .helpers import (
+    ApiError,
+    rm_tree_contents,
+    lieutenant_query,
+    sliding_window,
+    IndentedListDumper,
+)
 from .cluster import Cluster
 from .config import Config, Migration
 from .k8sobject import K8sObject
@@ -175,11 +181,31 @@ def _kapitan_029_030_difffunc(
     before_text: str, after_text: str, fromfile: str = "", tofile: str = ""
 ) -> tuple[Iterable[str], bool]:
 
+    diff_lines, _ = _ignore_yaml_formatting_difffunc(
+        before_text, after_text, fromfile, tofile
+    )
+
+    suppress_diff = not any(
+        _is_semantic_diff_kapitan_029_030(win)
+        for win in sliding_window(diff_lines[2:], 2)
+    )
+
+    return diff_lines, suppress_diff
+
+
+def _ignore_yaml_formatting_difffunc(
+    before_text: str, after_text: str, fromfile: str = "", tofile: str = ""
+) -> tuple[list[str], bool]:
+
     before_objs = sorted(yaml.safe_load_all(before_text), key=K8sObject)
-    before_sorted_lines = yaml.dump_all(before_objs).split("\n")
+    before_sorted_lines = yaml.dump_all(before_objs, Dumper=IndentedListDumper).split(
+        "\n"
+    )
 
     after_objs = sorted(yaml.safe_load_all(after_text), key=K8sObject)
-    after_sorted_lines = yaml.dump_all(after_objs).split("\n")
+    after_sorted_lines = yaml.dump_all(after_objs, Dumper=IndentedListDumper).split(
+        "\n"
+    )
 
     diff = difflib.unified_diff(
         before_sorted_lines,
@@ -188,13 +214,9 @@ def _kapitan_029_030_difffunc(
         fromfile=fromfile,
         tofile=tofile,
     )
-    diff_lines = list(diff)
-    suppress_diff = not any(
-        _is_semantic_diff_kapitan_029_030(win)
-        for win in sliding_window(diff_lines[2:], 2)
-    )
 
-    return diff_lines, suppress_diff
+    diff_lines = list(diff)
+    return diff_lines, len(diff_lines) == 0
 
 
 def update_catalog(cfg: Config, targets: Iterable[str], repo: GitRepo):
@@ -213,6 +235,9 @@ def update_catalog(cfg: Config, targets: Iterable[str], repo: GitRepo):
     if cfg.migration == Migration.KAP_029_030:
         click.echo(" > Smart diffing started... (this can take a while)")
         difftext, changed = repo.stage_all(diff_func=_kapitan_029_030_difffunc)
+    elif cfg.migration == Migration.IGNORE_YAML_FORMATTING:
+        click.echo(" > Smart diffing started... (this can take a while)")
+        difftext, changed = repo.stage_all(diff_func=_ignore_yaml_formatting_difffunc)
     else:
         difftext, changed = repo.stage_all()
     elapsed = time.time() - start
--- a/commodore/cli/catalog.py
+++ b/commodore/cli/catalog.py
@@ -94,10 +94,14 @@ def clean(config: Config, verbose):
     "--migration",
     help=(
         "Specify a migration that you expect to happen for the cluster catalog. "
-        + "Currently Commodore only knows the Kapitan 0.29 to 0.30 migration. "
+        + "Currently known are the Kapitan 0.29 to 0.30 migration and "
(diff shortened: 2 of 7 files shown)


bastjan:
@simu Every component with post process filters shows a diff. Sometimes huge ones (CRD).

Since `--migration kapitan-0.29-to-0.30` loads and dumps the YAML manifests it correctly ignores non semantic whitespace.

Should we add a more generic `--migration ignore-non-semantic-whitespace` migration?

Should we add a page Explanation/ Ignore Whitespace or just write release notes?

simu:
> @simu Every component with post process filters shows a diff. Sometimes huge ones (CRD).

That's expected.

> 
> Since `--migration kapitan-0.29-to-0.30` loads and dumps the YAML manifests it correctly ignores non semantic whitespace.
> 
> Should we add a more generic `--migration ignore-non-semantic-whitespace` migration?

I'd propose to refactor the whitespace bits of the kapitan-migration diff filter to be a generic `ignore-non-semantic-whitespace` filter

> 
> Should we add a page Explanation/ Ignore Whitespace or just write release notes?

I'll add something in the release notes, I don't think we need a page like we did for Kapitan 0.29 -> 0.30, as there's no inputs that need to be changed.

commodore/catalog.py
```diff
@@ -175,6 +175,24 @@ def _kapitan_029_030_difffunc(
     before_text: str, after_text: str, fromfile: str = "", tofile: str = ""
 ) -> tuple[Iterable[str], bool]:
 
+    [diff, _] = _ignore_non_semantic_whitespace_difffunc(
```
simu, line 178: No need to create a useless list to unpack the tuple return value, just using the following should work fine:

```suggestion
    diff, _ = _ignore_non_semantic_whitespace_difffunc(
```

commodore/catalog.py
```diff
@@ -175,6 +175,24 @@ def _kapitan_029_030_difffunc(
...
+        before_text, after_text, fromfile, tofile
+    )
+
+    diff_lines = list(diff)
+
```
simu, line 183: This is a noop, since we already call `list(diff_lines)` in `_ignore_non_semantic_whitespace_difffunc`

```suggestion
```

commodore/catalog.py
```diff
@@ -175,6 +175,24 @@ def _kapitan_029_030_difffunc(
...
+    return diff_lines, suppress_diff
+
+
+def _ignore_non_semantic_whitespace_difffunc(
+    before_text: str, after_text: str, fromfile: str = "", tofile: str = ""
+) -> tuple[Iterable[str], bool]:
+
     before_objs = sorted(yaml.safe_load_all(before_text), key=K8sObject)
     before_sorted_lines = yaml.dump_all(before_objs).split("\n")
```
simu, line 197: While this approach works to suppress non-semantic whitespace differences, it actually does a bunch more and will suppress reordering of objects in the output too. However, it doesn't look like Python's difflib has an option similar to `diff -b`, so this is probably the best we get.

Also, by using `yaml.dump_all()` without the custom dumper we use the wrong indentation for lists, which would generate inaccurate diffs if there's an actual change in a list combined with whitespace change.

simu (changes requested):
LGTM overall. Please update the help text for the catalog compile `--migration` option in https://github.com/projectsyn/commodore/blob/bc652134ad778470f12aadbf87a05448409608d2/commodore/cli/catalog.py#L92-L103. Also, it would be nice if we added the missing option in https://github.com/projectsyn/commodore/blob/master/docs/modules/ROOT/pages/reference/cli.adoc#catalog-compile.

commodore/catalog.py
bastjan: That would be my ruby background 🙈

commodore/catalog.py
bastjan: Mypy is unhappy without

commodore/catalog.py
bastjan: Maybe we should rename to `ignore-yaml-formatting-changes` or smth?

I'll use the same dumper

commodore/catalog.py
simu: ah, that would be because `_ignore_non_semantic_whitespace_difffunc()` indicates that it returns an `Iterable[str]`. Since we know that we return a list there, I'd just change that to `list[str]` in the return type.

commodore/catalog.py
simu: Renaming it to `ignore-yaml-formatting` is probably a good idea to not mislead users.

commodore/catalog.py
```diff
@@ -213,6 +235,9 @@ def update_catalog(cfg: Config, targets: Iterable[str], repo: GitRepo):
     if cfg.migration == Migration.KAP_029_030:
         click.echo(" > Smart diffing started... (this can take a while)")
         difftext, changed = repo.stage_all(diff_func=_kapitan_029_030_difffunc)
+    if cfg.migration == Migration.IGNORE_YAML_FORMATTING:
```
simu, line 238: I didn't see this yesterday evening, but this should be else if, to ensure that the three options (kapitan migration, yaml formatting migration, or regular diff) form a single logical block (it's not a bug per se, at the moment, since we still correctly execute the regular diff when neither migration is selected).

```suggestion
    elif cfg.migration == Migration.IGNORE_YAML_FORMATTING:
```

commodore/cli/catalog.py
```diff
@@ -94,10 +94,13 @@ def clean(config: Config, verbose):
...
         "Specify a migration that you expect to happen for the cluster catalog. "
-        + "Currently Commodore only knows the Kapitan 0.29 to 0.30 migration. "
+        + "Currently known are the Kapitan 0.29 to 0.30 migration and "
+        + "a generic migration ignoring all non-functional YAML formatting changes. "
         + "When the Kapitan 0.29 to 0.30 migration is selected, Commodore will suppress "
         + "noise (changing managed-by labels, and reordered objects) caused by the "
-        + "migration in the diff output."
+        + "migration in the diff output. "
+        + "When the ignore YAML formatting migration is selected, Commodore will suppress "
+        + "noise such as indentation and flow changes of lists or differences in string representation."
```
simu, line 103:
```suggestion
        + "noise such as reordered objects, indentation and flow changes of lists or differences in string representation."
```

docs/modules/ROOT/pages/reference/cli.adoc
```diff
@@ -75,6 +75,11 @@ Additionally, having Jsonnet and Kapitan dependency fetching enabled in local mo
 When you want to test adding a new component in local mode, you *must* run local mode with dependency fetching enabled at least once.
 After that, all the symlinks and dependencies which are required to compile the component will be present and you can disable dependency fetching.
 
+*-m, --migration*::
+  Specify a migration that you expect to happen for the cluster catalog.
+  Currently known are the Kapitan 0.29 to 0.30 migration and a generic migration ignoring all non-functional YAML formatting changes.
+  When the Kapitan 0.29 to 0.30 migration is selected, Commodore will suppress noise (changing managed-by labels, and reordered objects) caused by the migration in the diff output.
+  When the ignore YAML formatting migration is selected, Commodore will suppress noise such as indentation and flow changes of lists or differences in string representation.
```
simu, line 82:
```suggestion
  When the ignore YAML formatting migration is selected, Commodore will suppress noise such as reordered objects, indentation and flow changes of lists or differences in string representation.
```

simu (approved):

commodore/catalog.py
bastjan: 🤦‍♀️

Issue #627, opened 2022-09-15
kube-libsonnet is using soon deprecated API versions


megian:
## Context
With Kubernetes v1.25 a lot of API get [deprecated](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25). The latest version of kube-libsonnet still using it.
policy/v1beta1: https://github.com/bitnami-labs/kube-libsonnet/blob/23fe9f1c4905c174c90ea1b6681fb790b6715a6c/kube.libsonnet#L268
batch/v1beta1: https://github.com/bitnami-labs/kube-libsonnet/blob/23fe9f1c4905c174c90ea1b6681fb790b6715a6c/kube.libsonnet#L531

## Alternatives
* Drop the support of kube-libsonnet in Commodore

megian:
See bitnami-labs#75.

simu:
Possible options for making sure kube-libsonnet is compatible with K8s 1.25+ (no particular order):
* Hope that upstream accepts and merges https://github.com/bitnami-labs/kube-libsonnet/pull/76
* Vendor the library in Commodore (alongside `commodore.libjsonnet`) and take over maintenance for it
* Create a fork of https://github.com/bitnami-labs/kube-libsonnet in https://github.com/projectsyn, patch the fork and update Commodore to fetch the file from the projectsyn fork

simu:
After a slightly closer look. https://github.com/jsonnet-libs/k8s-libsonnet is not a drop-in replacement for kube-libsonnet, but actually looks much nicer (and more complete -- not too surprising since the base implementation is generated from the K8s API spec). However, we could probably engineer a migration path by providing a vendored kube-libsonnet which emits deprecation warnings informing users that they should switch to k8s-libsonnet.

megian:
The components projectsyn/component-kyverno and projectsyn/component-vault are affected and source of this issue.

PR #694, opened 2022-11-21
Vendor `kube-libsonnet`

bastjan:
Vendors `kube-libsonnet` and adds a depreciation notice that is currently not enabled by default.

### utils.libsonnet

Seems to be unused so I did not vendor it <https://github.com/search?q=%28%22utils.libsonnet%22+OR+%22utils.libjsonnet%22%29+AND+%28org%3Aprojectsyn+OR+org%3Aappuio+OR+org%3Avshn+%29&type=code>.

### Components with pinned `kube-libsonnet`

Only `vshn/cloudscale-metrics-collector` has added a [jsonnetfile.json](https://github.com/vshn/cloudscale-metrics-collector/blob/1fc786927d0cf4687ce3811fd1e2a6116bd82fa0/component/jsonnetfile.json#L7) with kube-libsonnet added. This was most likely by accident and I'll add a PR to remove it.

https://github.com/search?q=%22bitnami-labs%2Fkube-libsonnet%22+path%3A%2Fjsonnetfile%2F+AND+%28org%3Aprojectsyn+OR+org%3Aappuio+OR+org%3Avshn+%29&type=code

@simu and I decided that it is fine that `kube.libsonnet` can't be overridden by a components' Jsonnet Bundler file.

## Checklist

- [x] Update tests.
- [x] Link this PR to related issues. https://github.com/projectsyn/commodore/issues/627

### DIFF
--- a/tests/test_postprocess.py
+++ b/tests/test_postprocess.py
@@ -55,6 +55,10 @@ def _make_jsonnet_filter(tmp_path, ns, enabled=None, create_namespace=False):
                     kind: "Namespace",
                     metadata: {
                         name: params.namespace,
+                        // Annotations and labels are added to be consistent with the built in helm_namespace filter
+                        // which uses kube.libsonnet to create the namespace.
+                        annotations: { },
+                        labels: { name: params.namespace },
                     }
                 }
             }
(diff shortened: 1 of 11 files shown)


tests/test_postprocess.py
```diff
@@ -55,6 +55,8 @@ def _make_jsonnet_filter(tmp_path, ns, enabled=None, create_namespace=False):
                     kind: "Namespace",
                     metadata: {
                         name: params.namespace,
+                        annotations: { },
```
simu, line 58: Why do you add the empty `annotations: {}` here?

simu (approved):
LGTM overall

tests/test_postprocess.py
bastjan: The test now uses the real `kube.libjsonnet` which adds the `annotations: { }`. Also reminded me to test if the shim is still needed and removed it: https://github.com/projectsyn/commodore/pull/694/commits/d7e038abe822f1ed3576902e8350dc6276fcc167

bastjan: It's done to be consistent between the built in helm_namespace and the custom jsonnet filter.

simu: Ah fair enough, might be worth it to add a comment explaining this here

PR #707, opened 2022-12-12
Update Poetry to v1.3.1 in the Dockerfile

simu:
We were still using Poetry 1.1.13 in the build stage of the Dockerfile. With the latest changes to Poetry (cf. #706 ), the `poetry.lock` format has changed in an incompatible way. Since we can't control the Poetry version used by Renovate, we simply update to the latest Poetry version in the Dockerfile.

See also https://python-poetry.org/blog/announcing-poetry-1.3.0/

## Checklist


bastjan (approved):
Might be nice to get this version renovatable too

PR #746, opened 2023-03-01
Switch to Kapitan's new `literal` multi-line YAML string style

simu:
This renders multiline YAML strings with literal newlines, for example a field `data` which has three lines of text which contain "foo", "bar", and "baz" respectively will now be rendered as follows by Kapitan.

```yaml
data: |
  foo
  bar
  baz
```

This should improve readability of any manifests in the cluster catalog which contain large amounts of multi-line YAML strings, and should massively improve readability of catalog diffs for such manifests.

Follow-up to #720 

## Checklist


bastjan (approved):
🎉

PR #747, opened 2023-03-02
Only print a message when the OIDC login URL can't be opened automatically

simu:
Until now, we unconditionally printed a message containing the OIDC login URL whenever Commodore needed to fetch a new OIDC token. The `webbrowser.open()` function returns a boolean which indicates whether the URL was opened successfully.

This commit checks the return value from `webbrowser.open()` and only prints a message if the URL couldn't be opened. This reduces the chances of this message messing up the terminal output when autocompletion is used (cf. #722), and reduces noise in general.

Follow-up for #722 

## Checklist

### DIFF
--- a/commodore/login.py
+++ b/commodore/login.py
@@ -294,11 +294,12 @@ def login(config: Config):
         redirect_uri="http://localhost:18000",
         scope=["openid", "email", "profile"],
     )
-    print(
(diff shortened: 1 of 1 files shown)


glrf (approved):

simu:
Verified that message is printed as expected when running in Docker.

PR #750, opened 2023-03-07
Correctly merge instance parameters in `component compile`

simu:
Until now, `commodore component compile` didn't correctly merge parameters provided in the instance's parameters key when compiling a multi-instance-enabled component in it's instantiated form (using the `--alias` flag of `component compile`).

The root cause for this error was that we didn't use the same logic to generate the Kapitan target in `component compile` and `catalog compile` until now. The PR refactors `render_target()` to pull out the instance logic into a separate function `generate_target()` which we can use for `component compile` as well.

The new `generate_target()` function expects a list of classes to use for field `classes` in the resulting Kapitan target, but contains all the logic to render the contents of field `parameters` which is identical for `component compile` and `catalog compile`.

We do this refactoring, in order to not make `render_target()` more complex by introducing a new parameter to allow callers to provide additional classes to include. Instead we keep the interface of `render_target()` unchanged, but call `generate_target()` after constructing the list of classes required for `catalog compile`.

In `component compile`, we replace the hand-crafted target contents with a call to `generate_target()` using the same list of classes that we previously had in the hand-crafted target.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/cluster.py
+++ b/commodore/cluster.py
@@ -142,11 +142,50 @@ def read_cluster_and_tenant(inv: Inventory) -> tuple[str, str]:
     )
 
 
+def generate_target(
+    inv: Inventory,
+    target: str,
+    components: dict[str, Component],
+    classes: list[str],
+    component: str,
+):
+    """This function generates an object which is suitable to be marshalled into YAML as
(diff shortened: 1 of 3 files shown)


bastjan (approved):
I'm not 100% percent positive I understand the code. But the test looks correct.

PR #915, opened 2024-02-09
Ensure that cluster catalog is initialized with branch `master`

simu:
Until now, Commodore could fail to push the initial catalog for a cluster if the user's Git config sets `init.defaultBranch`. The
observed error is:

```
 > Commiting changes...
 > Pushing catalog to remote...
Error: Failed to push to the catalog repository: Git exited with status code 1
The error reported was:
  stderr: 'error: src refspec master does not match any
error: failed to push some refs to 'ssh://<catalog repo>''
```

This error is caused because GitPython falls back to `init.defaultBranch` when creating the catalog repo unless `initial_branch` is specified.

However, since Commodore unconditionally falls back to trying to push `master` when no default branch can be identified in the remote repo (e.g. empty catalog repo), the push then fails because we're trying to push branch `master` which doesn't exist locally.

This isn't an issue for `commodore component new` and `commodore package new` since we explicitly fall back to creating a `master` branch when initializing the worktree for the new dependency.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/gitrepo/__init__.py
+++ b/commodore/gitrepo/__init__.py
@@ -115,7 +115,7 @@ class GitRepo:
         if not force_init and targetdir.exists():
             self._repo = Repo(targetdir)
         else:
-            self._repo = Repo.init(targetdir, bare=bare)
+            self._repo = Repo.init(targetdir, bare=bare, initial_branch="master")
 
         if remote:
             self.remote = remote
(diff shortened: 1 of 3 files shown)


DebakelOrakel (approved):

simu:
Follow-up to rethink the handling of the default branch name in Commodore: #916

PR #1046, opened 2024-10-31
Use go-jsonnet in Commodore

simu:
We add go-jsonnet as a dependency, set `use_go_jsonnet=True` when calling `kapitan.targets.render_targets()` and replace our direct usages of jsonnet for rendering jsonnetfile.jsonnet for components and for postprocessing with go-jsonnet.

Additionally, we fix an oversight in the Jsonnet import callback used by the Commodore postprocessing where we should return a full path to the imported file but instead returned the filename only. With go-jsonnet this leads to an import loop for `lib/kube.libjsonnet` for the builtin `helm_namespace` filter. Interestingly, the same issue doesn't occur when using the same include in a Jsonnet postprocessing filter. Note that with C++ jsonnet we never had any issues with the previous import callback implementation. Additionally, note that Kapitan returns the full path in the import callback, cf. https://github.com/kapicorp/kapitan/blob/dfa5d33bd3e74db5fad84086c833402bc892da6f/kapitan/resources.py#L211-L257

## Checklist

- [x] Update tests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,15 @@
 FROM docker.io/python:3.11.5-slim-bullseye AS base
 
+ARG TARGETARCH
+ENV TARGETARCH=${TARGETARCH:-amd64}
+
 ENV HOME=/app
 
 WORKDIR ${HOME}
 
 FROM base AS builder
 
-ENV PATH=${PATH}:${HOME}/.local/bin
+ENV PATH=${PATH}:${HOME}/.local/bin:/usr/local/go/bin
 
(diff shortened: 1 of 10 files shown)


simu:
Needs https://github.com/appuio/component-openshift4-monitoring/pull/217 which is released in https://github.com/appuio/component-openshift4-monitoring/releases/tag/v6.5.1

haasad (approved):
LGTM, very excited to see the performance effects :rocket:

PR #1100, opened 2025-02-07
Fix component multi-version verification

simu:
We need to check whether a component instance that sets a custom version supports per-instance versions by checking the instance's target's parameters. The bootstrap target ("cluster") doesn't have any instance-specific parameters, but since we do the validation after setting up all instance targets, we can use the instance's target to read `_metadata` of the alias's version to properly validate that both the base component and the alias are multi-version aware.

Additionally, because `component compile` doesn't have the regular bootstrap target (and generating it just for alias validation seems a bit overkill), we make the `bootstrap_target` key an optional parameter for `verify_component_aliases()` and set it to the component that's being compiled for `component compile` and fall back to `config.inventory.boostrap_target` when it's not set.

We also adjust all tests that cover `verify_component_aliases()` to pass a Kapitan/reclass inventory dict instead of just a parameters dict and update the test inputs to set the alias `_metadata` in the  component parameters key instead of the instance parameters key to replicate the structure that we generate for `catalog compile`.

## Checklist

- [x] Update tests.

### DIFF
--- a/commodore/compile.py
+++ b/commodore/compile.py
@@ -236,7 +236,7 @@ def setup_compile_environment(config: Config) -> tuple[dict[str, Any], Iterable[
     create_component_library_aliases(config, cluster_parameters)
 
     # Verify that all aliased components support instantiation
-    config.verify_component_aliases(cluster_parameters)
+    config.verify_component_aliases(inventory)
     config.register_component_deprecations(cluster_parameters)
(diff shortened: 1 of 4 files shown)


haasad (approved):
I'm not super familiar with the code base, but the changes makes sense and match what is stated in the PR description

PR #1102, opened 2025-02-14
Make the backing `MultiDependency` for `Component` non-optional

simu:
This simplifies a lot of the implementation for `Component` and only surfaces in a minor way in `component_compile` where we may need to create a fake `MultiDependency` if we compile a component outside a Git repo.

Split out from #1101 

## Checklist

- [x] Update tests.


HappyTetrahedron (approved):
LGTM. Impressed with how much stuff can be deleted in the tests ^^

PR #1104, opened 2025-02-14
Correctly abort `catalog compile` for local changes in tenant repo

simu:
We incorrectly used `gr.has_local_branches()` and
`gr.is_ahead_of_remote()` when checking whether the tenant repo has local branches or is ahead of the tracking branch.

NOTE: we don't need to fetch first when checking if we're ahead of the tracking branch, since the cases we want to cover are 1) forgot to push and 2) push failed but got overlooked before running `catalog compile`.

## Checklist

- [ ] Update tests.


bastjan (approved):

haasad (approved):
:bug:

PR #1417, opened 2026-08-18
Run cleanup PR tag workflow on `merge_group` "destroyed" event

simu:
This should delete the container image tags created by merge queue CI runs.

## Checklist

- [x] Update tests.

### DIFF
--- a/.github/workflows/cleanup-pr-tag.yml
+++ b/.github/workflows/cleanup-pr-tag.yml
@@ -5,6 +5,9 @@ name: Delete closed PR container image tag
   pull_request:
     types:
       - closed
+  merge_group:
+    types:
+      - destroyed
 
 permissions:
   packages: write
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup-pr-tag.yml
```diff
@@ -14,7 +17,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Set image version for PR to branch name
+        if: github.event_name == 'pull_request'
         run: echo "VERSION=${GITHUB_HEAD_REF//\//-}" >> ${GITHUB_ENV}
+      - name: Set image version for merge queue to branch name
+        if: github.event_name == 'merge_group'
+        run: echo "VERSION=${GITHUB_REF//\//-}" >> ${GITHUB_ENV}
```
HappyTetrahedron, line 24: Why is this different between merge queues and PRs? iirc in the place where we create the tag names they're the same.

HappyTetrahedron: Ah nvm they're not.

HappyTetrahedron (approved):

.github/workflows/cleanup-pr-tag.yml
simu: not anymore, cf. https://github.com/projectsyn/commodore/blob/6f83b1e095ac7b40d267d65edf2c087b91cc4e06/.github/workflows/container-image.yml#L49-L54

I only found out that PRs and merge queues are different after you reviewed the other PR and it went through the merge queue for the first time.

PR #1419, opened 2026-08-18
Add envsubst dependency to commodore image

thobens:
This PR adds the `envsubst` dependency from https://github.com/a8m/envsubst to the docker image, so it can be used in the CI pipelines of commodore components. The following PRs are dependent on this change:

- https://github.com/projectsyn/component-capi-core/pull/1
- https://github.com/vshn/component-capi-provider-cloudscale/pull/1
- https://github.com/vshn/component-capi-provider-talos/pull/1

NOTE: Check if it would be better to integrate this dependency in the `commodore tools` command and whether this would be out of scope for ALDE-262

## Checklist

- [ ] Update tests.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,7 @@ FROM base AS builder
 ENV PATH=${PATH}:${HOME}/.local/bin:/usr/local/go/bin
 
 ARG POETRY_VERSION=2.4.1
+ARG ENVSUBST_VERSION=v1.4.3
 RUN apt-get update && apt-get install -y --no-install-recommends \
       build-essential \
       curl \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -14,13 +14,17 @@ FROM base AS builder
 ENV PATH=${PATH}:${HOME}/.local/bin:/usr/local/go/bin
 
 ARG POETRY_VERSION=2.4.1
+ARG ENVSUBST_VERSION=v1.4.3
```
simu, line 17: To get renovate support here, we need a new entry in `renovate.json`, cf. https://github.com/projectsyn/commodore/blob/cd33baa5c41d07babdeaa15532c54cd149347698/renovate.json#L51-L56

